PR #10, opened 2023-12-20
check route locale

Fbasham:
This prevents `/ennnnnn` or `frrrrrdsfdsffsdf/dfsdfsdf` from working for example.  It will result in a 404 if the locale isn't `en` or `fr` in the route.

### DIFF
--- a/frontend/app/utils/locale-utils.ts
+++ b/frontend/app/utils/locale-utils.ts
@@ -7,12 +7,13 @@ import { type UNSAFE_RouteModules as RouteModules } from '@remix-run/react';
  */
 export function getLocale(url: string) {
   const pathname = new URL(url).pathname;
+  const locale = pathname.split('/')?.[1];
 
-  if (pathname.startsWith('/en')) {
+  if (locale === 'en') {
     return 'en';
   }
 
-  if (pathname.startsWith('/fr')) {
+  if (locale === 'fr') {
     return 'fr';
   }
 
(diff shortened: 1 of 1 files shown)


frontend/app/utils/locale-utils.ts
```diff
@@ -7,12 +7,13 @@ import { type UNSAFE_RouteModules as RouteModules } from '@remix-run/react';
  */
 export function getLocale(url: string) {
   const pathname = new URL(url).pathname;
+  const locale = pathname.split("/")?.[1]
 
-  if (pathname.startsWith('/en')) {
+  if (locale==="en") {
```
sebastien-comeau, line 12: Nitpicking: add spaces before and after `===`. Ensure you have prettier plugin installed and confire your editor to auto-format on `.ts` and `.tsx`.

sebastien-comeau (changes requested):

frontend/app/utils/locale-utils.ts
```diff
@@ -7,12 +7,13 @@ import { type UNSAFE_RouteModules as RouteModules } from '@remix-run/react';
  */
 export function getLocale(url: string) {
   const pathname = new URL(url).pathname;
+  const locale = pathname.split('/')?.[1];
 
-  if (pathname.startsWith('/en')) {
+  if (locale === 'en') {
```
sebastien-comeau, line 12: Is locale case always be in lowercase? If someone type `/EN/home` is should works. Please check if `pathname` is always in lowercase.

gregory-j-baker (approved):

frontend/app/utils/locale-utils.ts
gregory-j-baker: `/EN` returns a 404, as expected

frontend/app/utils/locale-utils.ts
gregory-j-baker: Auto-formatters are satan's spawn

nicholas-ly (approved):

frontend/app/utils/locale-utils.ts
sebastien-comeau: Its depends on the platform. Remix is case sensitive and Next.js is not based on the document sites.

sebastien-comeau (approved):

PR #13, opened 2023-12-21
Add security headers

gregory-j-baker:
See <https://cheatsheetseries.owasp.org/cheatsheets/HTTP_Headers_Cheat_Sheet.html>


sebastien-comeau (approved):
Looks good to me.

PR #17, opened 2023-12-21
Switch to cookie/query language toggle

gregory-j-baker:


sebastien-comeau:
Updated react-i18next to v14.0.0. That upgrade have a fix for your typescript issue.

See changelog: https://github.com/i18next/react-i18next/blob/master/CHANGELOG.md

frontend/app/components/language-switcher.tsx
```diff
@@ -14,11 +14,10 @@ export function LanguageSwitcher({ ...props }: LanguageSwitcherProps) {
   const { pathname } = useLocation();
   const { i18n, t } = useTranslation('common');
 
-  const altLang = i18n.language === 'fr' ? 'en' : 'fr';
-  const linkPath = pathname.replace(i18n.language, altLang);
+  const to = `${pathname}?lang=${i18n.language === 'fr' ? 'en' : 'fr'}`;
 
   return (
-    <Link {...props} to={linkPath} onClick={() => i18n.changeLanguage(altLang)}>
+    <Link to={to} reloadDocument {...props}>
```
sebastien-comeau, line 20: Spread first than override.

```suggestion
    <Link {...props} to={to} reloadDocument={true}>
```

frontend/app/utils/locale-utils.server.ts
```diff
@@ -1,8 +1,25 @@
...
+import { parse, serialize } from 'cookie';
 import { type Namespace, createInstance } from 'i18next';
 import I18NexFsBackend from 'i18next-fs-backend';
 import { resolve } from 'node:path';
 
-import { getLocale } from '~/utils/locale-utils';
+export function createLangCookie() {
+  const cookieName = '_gc_lang';
+
+  const langCookie = createCookie(cookieName, {
```
sebastien-comeau, line 11: We shouldn't use `createCookie` because overriding `parse` and `serialize` doesn't work as you thought. `parseOptions` and `serializeOptions` are always underfined based on my tests. We should use `parse` and `serialize` func from `cookie` package for that specific cookie because it needs to work differently then Remix's cookies. The value of the cookie should be base64 encoded in our case.

gregory-j-baker: Acknowledged; will fix.

frontend/app/components/language-switcher.tsx
gregory-j-baker: In this case it doesn't matter because of `type LanguageSwitcherProps = Omit<LinkProps, 'to' | 'reloadDocument'>`, but I will change it anyway for consistency with other code that might follow suit.

gregory-j-baker: [Fixed](https://github.com/DTS-STN/canadian-dental-care-plan/pull/17/files#diff-5a4ca5b528aaf8411c9cb8f6c6ba138f324ef5ae6d39537f8c6a6c2e204e78a9R20)

frontend/app/utils/locale-utils.server.ts
gregory-j-baker: [Fixed](https://github.com/DTS-STN/canadian-dental-care-plan/pull/17/files#diff-73af98351d32bbf3e05728a23721407c32230588067302df2571f014a28bc9d9R21-R30)

No longer making a call to Remix's `createCookie(..)` function. Instead, I am directly returning a `Cookie` type.

nicholas-ly (approved):

PR #19, opened 2023-12-22
add playwright e2e testing

Fbasham:
Add e2e testing capability. Based on Kent C Dodds and the remix-run examples, playwright is preferred over cypress.  I know for retirement hub we did this in a separate folder; unsure if that's how that should be done for this project.  Take a look and see what you think.  Maybe cypress is still the way to go too.

I followed the config here:

https://github.com/remix-run/examples/tree/main/playwright
https://github.com/kentcdodds/kentcdodds.com

### DIFF
--- a/frontend/.gitignore
+++ b/frontend/.gitignore
@@ -4,3 +4,7 @@ node_modules
 /build
 /public/build
 .env
+/test-results/
+/playwright-report/
+/blob-report/
+/playwright/.cache/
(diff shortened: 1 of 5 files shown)


gregory-j-baker (approved):
Seems good to me. There are some extra testing features we can pull in from KCD's Remix epic stack in a future PR, if we feel like it: <https://github.com/epicweb-dev/epic-stack>

PR #20, opened 2024-01-04
Add working `dockerfile` 

gregory-j-baker:
As per title.


nicholas-ly (approved):
Builds and runs as expected on my PC - approving

PR #30, opened 2024-01-08
TypeScript definitions for i18next

sebastien-comeau:
> i18next has embedded type definitions. If you want to enhance IDE Experience and prevent errors (such as type coercion), you should follow the instructions below in order to get the t function fully-type safe (keys and return type).

https://www.i18next.com/overview/typescript

### DIFF
--- a/frontend/app/types.d.ts
+++ b/frontend/app/types.d.ts
@@ -1,0 +1,5 @@
+import { type Namespace } from 'i18next';
+
+export interface RouteHandle extends Record<string, unknown> {
+  i18nNamespaces?: Namespace;
+}
--- a/frontend/app/routes/_gcweb-app.$.tsx
+++ b/frontend/app/routes/_gcweb-app.$.tsx
@@ -3,11 +3,12 @@ import { Link } from '@remix-run/react';
 
 import { Trans, useTranslation } from 'react-i18next';
 
+import { type RouteHandle } from '~/types';
 import { getFixedT } from '~/utils/locale-utils.server';
 
 export const handle = {
   i18nNamespaces: ['gcweb'],
-};
+} satisfies RouteHandle;
 
 export const loader = async ({ request }: LoaderFunctionArgs) => {
   const t = await getFixedT(request, handle.i18nNamespaces);
--- a/frontend/app/routes/_gcweb-app._index.tsx
+++ b/frontend/app/routes/_gcweb-app._index.tsx
@@ -3,11 +3,12 @@ import { Link } from '@remix-run/react';
 
 import { useTranslation } from 'react-i18next';
 
+import { type RouteHandle } from '~/types';
 import { getFixedT } from '~/utils/locale-utils.server';
 
 export const handle = {
   i18nNamespaces: ['common'],
-};
+} satisfies RouteHandle;
 
 export const loader = async ({ request }: LoaderFunctionArgs) => {
   const t = await getFixedT(request, handle.i18nNamespaces);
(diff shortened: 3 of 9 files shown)


frontend/app/types/cdcp.d.ts
```diff
@@ -0,0 +1,5 @@
+import { type RouteHandle as RemixRouteHandle } from '@remix-run/react/dist/routeModules';
+
+import { type Namespace } from 'i18next';
+
+export type RouteHandle = RemixRouteHandle & { i18nNamespaces: Namespace };
```
gregory-j-baker, line 5: I don't know if this makes sense. The data in a route handle depends on the needs of the route.

Not all routes will need an i18n namespace (ie: API routes), and some routes may need to emit different optional data (ie: breadcrumb data).

frontend/app/types/cdcp.d.ts
gregory-j-baker, line 1: If we are going to have global application-scoped types, why not just create an `app/types.d.ts` file? Do we need to have a subfolder for this?

frontend/@types/i18next.d.ts
gregory-j-baker, line 1: Why not move this code into `remix.env.d.ts`? I just tested this and it works the same.

image

frontend/app/types/cdcp.d.ts
sebastien-comeau: will fix

frontend/@types/i18next.d.ts
gregory-j-baker: ex:

``` typescript
import type common from './public/locales/en/common.json';
import type gcweb from './public/locales/en/gcweb.json';

/**
 * @see https://www.i18next.com/overview/typescript
 */
declare module 'i18next' {
  interface CustomTypeOptions {
    resources: {
      common: typeof common;
      gcweb: typeof gcweb;
    };
  }
}

/// <reference types="@remix-run/dev" />
/// <reference types="@remix-run/node" />
```

frontend/app/types/cdcp.d.ts
sebastien-comeau: fixed!

frontend/@types/i18next.d.ts
sebastien-comeau: > Note that the types referenced in remix.env.d.ts will depend on which environment you're running your app in. For example, there are different globals available in Cloudflare.

https://remix.run/docs/en/main/guides/typescript

gregory-j-baker: Your comment doesn't make sense. If we decide to deploy to a non-node environment, then we'll change the type reference.

My point is that we have a file for defining types that aren't included in the standard runtime, so why not use it to define the i18next types?

sebastien-comeau: Fixed

frontend/app/types/cdcp.d.ts
sebastien-comeau: Fixed! :)

frontend/app/types.d.ts
```diff
@@ -0,0 +1,5 @@
+import { type Namespace } from 'i18next';
+
+export interface RouteHandle extends Record<string, any> {
```
gregory-j-baker, line 3:
```suggestion
export interface RouteHandle extends Record<string, unknown> {
```

gregory-j-baker (changes requested):

frontend/app/types.d.ts
sebastien-comeau: Fixed

gregory-j-baker (approved):

PR #31, opened 2024-01-08
Prepwork for future PR

gregory-j-baker:
Updating types in the i18n `getNamespaces()` function in preparation for an upcoming overhaul to the page metadata subsystem (page title, page id, build info, etc).

(also fixed broken types)

### DIFF
--- a/frontend/app/utils/locale-utils.ts
+++ b/frontend/app/utils/locale-utils.ts
@@ -7,27 +7,18 @@ import { type RouteHandle } from '~/types';
 
 /**
  * Returns all namespaces required by the given routes by examining the route's i18nNamespaces handle property.
- *
- * @param {RouteModules} routeModules - Object containing route modules
  * @see https://remix.run/docs/en/main/route/handle
  */
-export function getNamespaces(routeModules: Record<string, { handle?: unknown }>) {
-  const namespaces = new Set(
-    Object.values(routeModules)
-      .map((route) => route.handle)
-      .filter((handle): handle is RouteHandle => handle !== undefined)
-      .map((handle) => handle.i18nNamespaces)
-      .filter((i18nNamespaces): i18nNamespaces is Namespace => i18nNamespaces !== undefined)
-      .flatMap((i18nNamespaces) => i18nNamespaces),
-  );
-
-  return [...namespaces];
+export function getNamespaces(routes: Array<{ handle?: unknown }>) {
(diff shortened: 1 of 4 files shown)


frontend/app/utils/locale-utils.ts
```diff
@@ -7,11 +9,9 @@ import { type RouteHandle } from '~/types';
 
 /**
  * Returns all namespaces required by the given routes by examining the route's i18nNamespaces handle property.
- *
- * @param {RouteModules} routeModules - Object containing route modules
  * @see https://remix.run/docs/en/main/route/handle
  */
-export function getNamespaces(routeModules: Record<string, { handle?: unknown }>) {
+export function getNamespaces(routeModules: RouteModules | UIMatch[]) {
```
sebastien-comeau, line 14: Both types match the `Record<string, { handle?: unknown }>`, right? I don't think you need to specify multiple type if the are conform to `Record<string, { handle?: unknown }>`, right? I ease the testing of getNamespaces since you only need to provide `Record<string, { handle?: unknown }>`.

gregory-j-baker: `UIMatch[]` doesn't satisfy `Record<string, { handle?: unknown}>`, which is why I added it.

Try doing this:

``` typescript
const matches = useMatches();
const namespaces = getNamespaces(matches);
```

gregory-j-baker: In both cases (record and UIMatch), `Object.values()` gives me an object that contains a `handle`, so even though the types are a bit weird, it works.

I'm open to suggestions on how to make it better. My other thought was to just create two methods: `getNamespacesFromRouteModules()` and `getNamespacesFromUIMatches()`.

I'm not sure if that's an improvement, though.

sebastien-comeau: We don't care about the Record and his key. Use an array of object with handle prop instead.

gregory-j-baker: k.. lemme try to implement that.

gregory-j-baker: [Fixed](https://github.com/DTS-STN/canadian-dental-care-plan/pull/31/files#diff-597723d1c49211170cac306c23d89375db5ca5226555f2d096ff8d4f9dc44cfc)

sebastien-comeau (dismissed):

sebastien-comeau (approved):

PR #34, opened 2024-01-09
Environment variable validation

gregory-j-baker:
Server and client environment variable validation via zod.


frontend/app/utils/env.server.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+export function getEnv<T>(key: ServerEnvKey) {
+  const result = serverEnv.safeParse(process.env);
+
+  if (!result.success) {
+    throw new Error(`Invalid application configuration: ${result.error}`);
+  }
+
+  return result.data[key] as T;
+}
```
sebastien-comeau, line 40: Return the parsed config and let the consumer use the config key with destructuring assignment if he need a specific key... that way everything will be typed.

```suggestion
export function getEnv() {
  const result = serverEnv.safeParse(process.env);

  if (!result.success) {
    throw new Error(`Invalid application configuration: ${result.error}`);
  }

  return result;
}
```

gregory-j-baker: When I remove the `<T>` generic, this happens...

This should be a `string`:

image

These values' types should be tighter:

image

gregory-j-baker: I fixed this as per your suggestions. Please re-review.

frontend/app/utils/env.server.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  if (!result.success) {
+    throw new Error(`Invalid application configuration: ${result.error}`);
+  }
+
+  return result.data;
+}
+
+export function getClientEnv() {
+  const result = clientEnv.safeParse(process.env);
```
sebastien-comeau, line 40: You can't read `process.env` from the client.

gregory-j-baker: This is a `.server.ts` file.. It's only ever used server-side. The function name is `getClientEnv()` because it specifically returns client-safe variables (to be used in a server-side loader function).

sebastien-comeau: ok, so how do you read `env` from the client since it would only read `window.env`?

sebastien-comeau: Ok, I think I got it... it's confusing but I think I understand the flow.

sebastien-comeau (approved):

PR #39, opened 2024-01-09
Fix lang switcher url search params

sebastien-comeau:
Add `lang` search param in the existint search params for the current url.

### DIFF
--- a/frontend/app/components/language-switcher.tsx
+++ b/frontend/app/components/language-switcher.tsx
@@ -1,4 +1,4 @@
-import { Link, type LinkProps, useHref } from '@remix-run/react';
+import { Link, type LinkProps, useSearchParams } from '@remix-run/react';
 
 import { useTranslation } from 'react-i18next';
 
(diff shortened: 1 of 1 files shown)


gregory-j-baker (commented):
This is better than before, but still doesn't handle URL frags correctly.

Steps to reproduce:
- run application locally
- navigate to <http://localhost:3000/about#wb-cont>
- toggle language

notice the URL frag is lost

sebastien-comeau:
> This is better than before, but still doesn't handle URL frags correctly.
> 
> Steps to reproduce:
> 
> * run application locally
> * navigate to http://localhost:3000/about#wb-cont
> * toggle language
> 
> notice the URL frag is lost

URL fragement or harsh are not sent to the server and it cause hydration issue.

```
react-dom.development.js:86 Warning: Prop href did not match. Server: "/about?test=moi&lang=en" Client: "/about?test=moi&lang=en#wb-cont"
```

PR #42, opened 2024-01-10
add mock service worker

Fbasham:
Stub out mock service worker to eventually mock API calls.

Following this repo for reference:  https://github.com/mswjs/examples/tree/main/examples/with-remix


sebastien-comeau (approved):
Tested locally and seems to work fine. First step done! :)

PR #59, opened 2024-01-12
Bump dependencies

gregory-j-baker:
Remix v2.4.x → v2.5.0.


sebastien-comeau (dismissed):

frontend/app/utils/locale-utils.ts
```diff
@@ -25,9 +25,9 @@ export function getAltLanguage(language: string) {
  * Returns all namespaces required by the given routes by examining the route's i18nNamespaces handle property.
  * @see https://remix.run/docs/en/main/route/handle
  */
-export function getNamespaces(routes: Array<{ handle?: unknown }>) {
+export function getNamespaces(routes: Array<Partial<{ handle: unknown }> | undefined>) {
```
sebastien-comeau, line 28: Don't use `Partial` unless you are using an already defined interfaces or types (`RouteModule` or `EntryRouteModule` or `UIMatch`) and want to make their properties all optional. In our case we are defining the type explicitly.

```suggestion
export function getNamespaces(routes: Array<{ handle?: unknown } | undefined>) {
```

gregory-j-baker: [Fixed](https://github.com/DTS-STN/canadian-dental-care-plan/pull/59/files#diff-597723d1c49211170cac306c23d89375db5ca5226555f2d096ff8d4f9dc44cfcR28)

sebastien-comeau (approved):

PR #60, opened 2024-01-12
Decouple logging configuration

gregory-j-baker:
Since many components rely on logging to be fully configured, its configuration needs to be decoupled from other components.

Previously, logging was configured using the `getEnv()` function, but if that function threw an error, the logger would remain un-initialized and nothing would get logged.


frontend/app/utils/logging.server.ts
```diff
@@ -1,12 +1,18 @@
...
+// ['error', 'warn', 'info', 'verbose', 'debug', 'silly']
+const logLevels = Object.keys(winston.config.npm.levels);
 
-export const getLogger = (category: string): Logger => {
-  const { LOG_LEVEL } = getEnv();
+const logLevel = z
+  .string()
+  .refine((val) => logLevels.includes(val), { message: 'Invalid log level', path: ['process.env.LOG_LEVEL'] })
+  .default('info')
+  .parse(process.env.LOG_LEVEL);
```
sebastien-comeau, line 11: Is it working of the logger is initialised on the client since it doesn't have access to `process.env.LOG_LEVEL`?

sebastien-comeau: Will approve since `getLogger` is only called in server files for now. Keep in mind that we might want to log on the client later. :)

sebastien-comeau (approved):

PR #62, opened 2024-01-12
refactor(frontend): use pop() to returns last element

sebastien-comeau:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/pop

### DIFF
--- a/frontend/app/root.tsx
+++ b/frontend/app/root.tsx
@@ -58,7 +58,7 @@ function PageTitle() {
     .filter((handle): handle is RouteHandle => handle !== undefined)
     .filter((handle) => handle.pageTitlei18nKey !== undefined)
     .map((handle) => handle.pageTitlei18nKey)
-    .reduce((last, curr) => curr ?? last);
+    .pop();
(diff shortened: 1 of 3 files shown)


frontend/app/root.tsx
```diff
@@ -58,7 +58,7 @@ function PageTitle() {
     .filter((handle): handle is RouteHandle => handle !== undefined)
     .filter((handle) => handle.pageTitlei18nKey !== undefined)
     .map((handle) => handle.pageTitlei18nKey)
-    .reduce((last, curr) => curr ?? last);
+    .pop();
```
gregory-j-baker, line 61: This isn't functionally the same. If the last element is `undefined`, then `pop()` will return `undefined`.

`.reduce((last, curr) => curr ?? last)` will return the last non-`undefined` element, unless they're all undefined.

sebastien-comeau: You're right. Sorry for my confusion. Thanks :)

PR #63, opened 2024-01-12
2761 Setup end-to-end accessibility testing

shawn320:
[2761 Setup end-to-end accessibility testing](https://dev.azure.com/DTS-STN/Canada%20Dental%20Care%20Plan/_workitems/edit/2761/)

Add a11y check for homepage.


gregory-j-baker (commented):
Looks okay to me but it won't merge until the tests pass. I'm currently getting a fail for `[chromium] › example.spec.ts:13:3 › homepage › should not have any automatically detectable accessibility issues`

frontend/app/routes/_gcweb-app.tsx
```diff
@@ -207,7 +207,7 @@ function Breadcrumbs() {
   }
 
   return (
-    <nav id="wb-bc" property="breadcrumb">
+    <nav id="wb-bc" property="breadcrumb" aria-label="breadcrumbs">
```
sebastien-comeau, line 210: I don't think we have to add `aria-label` since the `h2` is the next element but if we need to to then add an `id` attribute to `h2` and refer it with `aria-labelledby` instead. That way the label will comes from `h2` content and will be translated.

frontend/app/routes/_gcweb-app.tsx
```diff
@@ -170,10 +170,10 @@ function PageFooter() {
 
   return (
     <footer id="wb-info">
-      <h2 className="wb-inv">{t('gcweb:footer.about-site')}</h2>
+      <h2 id='footer-heading' className="wb-inv">{t('gcweb:footer.about-site')}</h2>
       <div className="gc-sub-footer">
         <div className="d-flex align-items-center container">
-          <nav>
+          <nav aria-labelledby='footer-heading'>
```
sebastien-comeau, line 176: I think the `aria-labelledby` element's id should be the next `h3` and not the closest `h2`.

sebastien-comeau (approved):
Well done. :)

PR #65, opened 2024-01-12
Move page titles to loader

gregory-j-baker:


frontend/app/utils/route-utils.ts
```diff
@@ -0,0 +1,10 @@
+import { useMatches } from '@remix-run/react';
+
+export function usePageTitle() {
+  return useMatches()
+    .map((match) => match.data)
+    .filter((data): data is { pageTitle?: string } => data !== undefined && data !== null)
+    .filter((data) => data?.pageTitle !== undefined && data?.pageTitle !== null)
+    .map((data) => data.pageTitle)
+    .reduce((last, curr) => curr ?? last, undefined);
+}
```
sebastien-comeau, line 10: Use `Zod` for the object validation and infer PageTitle object type with it.

```suggestion
export function usePageTitle() {
  return useMatches()
    .map(({ data }) => data as any)
    .map(({ data }) => {
      const result = pageTitle.safeParse(data);
      if (result.success) return result.data;
      return undefined;
    })
    .filter((item): item is PageTitle => item !== undefined)
    .map(({pageTitle}) => pageTitle)
    .reduce((last, curr) => curr ?? last, undefined); 
}
```

gregory-j-baker: Will do, friend.. hang tight.

gregory-j-baker: This guy.. who does he think he is?

gregory-j-baker: Oops wrong window.

gregory-j-baker: [Fixed](https://github.com/DTS-STN/canadian-dental-care-plan/pull/65/files#diff-fa420e73484360a54e4949f50d2e0c50929565e217a8d566ad69a266c692abc0R5-R12)

frontend/app/utils/route-utils.ts
```diff
@@ -0,0 +1,12 @@
...
+
+import { z } from 'zod';
+
+const pageTitleDataSchema = z.object({ pageTitle: z.string() });
+
+export function usePageTitle() {
+  return useMatches()
+    .map(({ data }) => pageTitleDataSchema.safeParse(data))
+    .map((result) => (result.success ? result.data.pageTitle : undefined))
+    .reduce((last, curr) => curr ?? last, undefined);
```
sebastien-comeau, line 11: If you set `ES2023` instead of `ES2022` in tsconfig.json lib option you will gain access to [findLast](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findLast)
```suggestion
  return useMatches().findLast(({ data }) => {
    const result = pageTitleDataSchema.safeParse(data);
    return result.success ? result.data.pageTitle : undefined;
```

gregory-j-baker: We can do this in a future PR and change all reducers to `.findLast()` in one commit.

`/wontfix`

sebastien-comeau: Alright.

sebastien-comeau (approved):

PR #67, opened 2024-01-12
refactor(frontend): Move breadcrumb data to loader

sebastien-comeau:


frontend/app/utils/route-utils.ts
```diff
@@ -4,9 +4,29 @@ import { z } from 'zod';
...
+    z.object({
+      label: z.string(),
+      to: z.string().optional(),
+    }),
+  ),
+});
+
+export type BreadcrumbDataSchema = z.infer<typeof breadcrumbDataSchema>;
+
+export function useBreadcrumb() {
```
gregory-j-baker, line 27: **Nitpick:** should this be `useBreadcrumbs()`?

gregory-j-baker (approved):

frontend/app/utils/route-utils.ts
sebastien-comeau: 🤷

The correct term is "breadcrumb" without an "s." In the context of a website, a breadcrumb is a navigational aid that helps users understand their location on a site and easily navigate back to previous pages. It's called a breadcrumb trail, and the term is singular in this context. So, you would refer to it as "breadcrumb" when discussing website navigation.

🤷

sebastien-comeau: I guess you're right.

> If your React hook returns an array that may contain zero or multiple breadcrumb items, you might want to use the plural form "breadcrumbs" when referring to the array as a whole. This is because the array could potentially represent multiple instances of the breadcrumb.

PR #71, opened 2024-01-15
refactoring(frontend): add getTypedI18nNamespaces utility function

sebastien-comeau:
Using type `Namespace` to declare `i18nNamespaces` is not wrong but gives unwanted issue when using that variable with `getFixedT` or `useTranslation` because they cannot infer the actual values assignement to suggest or validate i18n keys to use or be used.

image

One way to fix that is to declare the variable as `const i18nNamespaces = ['common'] as const satisfies Namespace;`.  The infered type is `readonly ["common"]` which is good but then you need to do it everywhere.

The `getTypedI18nNamespaces` utility function that I created will take care of returning `readonly ["common"]`.

Only keys for `common` namespace will be returned.

image

### DIFF
--- a/frontend/app/routes/_gcweb-app.$.tsx
+++ b/frontend/app/routes/_gcweb-app.$.tsx
@@ -1,22 +1,22 @@
 import { type LoaderFunctionArgs, json } from '@remix-run/node';
 import { Link } from '@remix-run/react';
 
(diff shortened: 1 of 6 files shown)


gregory-j-baker (approved):
Thanks, I hate it.

PR #75, opened 2024-01-16
feature(i18n): configure i18next option

sebastien-comeau:
Set `appendNamespaceToMissingKey` option to `true`.
https://www.i18next.com/overview/configuration-options#missing-keys


gregory-j-baker:
Does this allow us to omit the namespace when doing `t(myKey)`?

sebastien-comeau:
> Does this allow us to omit the namespace when doing `t(myKey)`?

It adds the NS to the missing key... so except to return the key only `about.breadcrumbs.home` it returns `common:about.breadcrumbs.home`

PR #77, opened 2024-01-16
feateure(i18n): use i18next parsed keys

sebastien-comeau:
Utilize i18next parsed keys alongside `LoaderFunctionData` types to retrieve translation keys from the server rather than pre-translated text. The issue with pre-translated text is that upon retrieval, you need to invoke `revalidator.revalidate()` when the language changes to obtain the most up-to-date translated loader content. This poses a problem for users, as they lose form action validation if the form was submitted beforehand.


frontend/app/types.d.ts
```diff
@@ -2,7 +2,7 @@ import { type ParseKeys } from 'i18next';
 
 import type common from '../public/locales/en/common.json';
 import type gcweb from '../public/locales/en/gcweb.json';
-import { type Breadcrumbs, type BuildInfo, type I18nNamespaces, type PageIdentifier, type PageTitle } from './utils/route-utils';
+import { type Breadcrumbs, type BuildInfo, type I18nNamespaces, type PageIdentifier, type PageTitleI18nKey } from './utils/route-utils';
```
gregory-j-baker, line 5:
```suggestion
import { type Breadcrumbs, type BuildInfo, type I18nNamespaces, type PageIdentifier, type PageTitleI18nKey } from '~/utils/route-utils';
```

gregory-j-baker: 🤷 --- I wonder if we can set up an eslint rule to enforce this.

sebastien-comeau: Fixed

gregory-j-baker (approved):

gregory-j-baker (approved):

PR #78, opened 2024-01-16
Move cached files to `node_modules/.cache/` to keep root clean

gregory-j-baker:
As per title.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -11,9 +11,9 @@
     "dev:server": "cross-env NODE_ENV=development npm run-script build:server -- --watch",
     "start": "cross-env NODE_ENV=production node ./build/server.js ./build/index.js",
     "format": "prettier --write ./",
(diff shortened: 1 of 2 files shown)


sebastien-comeau (commented):
I don't see the need for it and does it make sense to have `.cache` inside `node_modules`?

gregory-j-baker:
> I don't see the need for it and does it make sense to have `.cache` inside `node_modules`?

It's not needed, but it's also not unconventional. Most of the official Remix templates use it, as does KCD's epic stack.

- <https://github.com/remix-run/blues-stack/blob/main/remix.config.js#L3>
- <https://github.com/remix-run/indie-stack/blob/main/remix.config.js#L3>
- <https://github.com/epicweb-dev/epic-stack/blob/main/remix.config.js#L7>

sebastien-comeau (approved):

PR #79, opened 2024-01-16
add unit test for language switcher

Fbasham:
This PR demonstrates one method for unit testing components.  We need `createRemixStub` because the component relies on a remix context.  From what I'm able to discern from others online, mocking is needed when a component internally relies on helper functions/utils.  Without these mocks, the tests will not be able to run.


frontend/__tests__/components/language-switcher.test.tsx
```diff
@@ -0,0 +1,46 @@
+import { render, screen, waitFor } from '@testing-library/react';
+
+import { createRemixStub } from '@remix-run/testing';
+
+import { axe, toHaveNoViolations } from 'jest-axe';
```
gregory-j-baker, line 5:
```suggestion
import { axe, toHaveNoViolations } from 'jest-axe';
import { describe, expect, it } from 'vitest';
```

We want to ensure we're using vitest everywhere instead of jest. The test still works with this change.

frontend/__tests__/components/language-switcher.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+  return {
+    getAltLanguage,
+  };
+});
+
+describe('Language Switcher', () => {
+  it('Should render and not have a11y issues', async () => {
+    const RemixStub = createRemixStub([
+      {
+        Component: () => <LanguageSwitcher>English</LanguageSwitcher>,
```
gregory-j-baker, line 33: **Nitpick:**

```suggestion
        Component: () => <LanguageSwitcher>Français</LanguageSwitcher>,
```
<br/>

**Reasoning:** it could be confusing to future devs to see `English` with the toggle href being `?lang=en`. It confused me for a minute.

frontend/__tests__/setup-test-env.ts
```diff
@@ -1,7 +1,13 @@
...
 // Workaround: For some reason FormData is not set to jsdom's by default
 const jsdom = new JSDOM(`<!doctype html>`);
 const { FormData } = jsdom.window;
 window.FormData = FormData;
 global.FormData = FormData;
+
+afterEach(() => {
+  vi.clearAllMocks();
+  vi.resetAllMocks();
+});
```
gregory-j-baker, line 13: 1. doesn't `resetAllMocks()` also clear all mocks?
2. While I'm not totally opposed to this, wouldn't it make more sense for the creator of the mocks to also control their lifecycles? What if a test needed a mock's state to persist across multiple tests? (and yes, I acknowledge this is probably bad practice)

gregory-j-baker (changes requested):

frontend/__tests__/components/language-switcher.test.tsx
Fbasham: I believe you're suggesting adding an explicit import from `vitest`?  If so, added.

gregory-j-baker (dismissed):
A+++++++ would review again!

gregory-j-baker (approved):

PR #80, opened 2024-01-16
Use route handle for common page data

sebastien-comeau:
As per title.

### DIFF
--- a/frontend/app/routes/_gcweb-app.$.tsx
+++ b/frontend/app/routes/_gcweb-app.$.tsx
@@ -7,16 +7,15 @@ import { getTypedI18nNamespaces } from '~/utils/locale-utils';
 
 const i18nNamespaces = getTypedI18nNamespaces('gcweb');
 
-export const loader = async ({ request }: LoaderFunctionArgs) => {
-  return json(
-    {
-      i18nNamespaces,
-      pageIdentifier: 'CDCP-0404',
-      pageTitleI18nKey: 'gcweb:not-found.page-title',
-    } as const satisfies LoaderFunctionData,
-    { status: 404 },
-  );
-};
+export const handle = {
+  i18nNamespaces,
+  pageIdentifier: 'CDCP-0404',
+  pageTitleI18nKey: 'gcweb:not-found.page-title',
+} as const satisfies RouteHandleData;
+
+export async function loader({ request }: LoaderFunctionArgs) {
+  return json(null, { status: 404 });
+}
 
 export default function NotFound() {
   const { t } = useTranslation(i18nNamespaces);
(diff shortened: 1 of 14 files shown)


frontend/app/routes/_gcweb-app.$.tsx
```diff
@@ -7,15 +7,14 @@ import { getTypedI18nNamespaces } from '~/utils/locale-utils';
...
-  );
+export const handle = {
+  i18nNamespaces,
+  pageIdentifier: 'CDCP-0404',
+  pageTitleI18nKey: 'gcweb:not-found.page-title',
+} as const satisfies RouteHandleData;
+
+export const loader = async () => {
+  return json(null, { status: 404 });
 };
```
gregory-j-baker, line 18:
```suggestion
export async function loader({ request }: LoaderFunctionArgs) {
  return json(null, { status: 404 });
};
```

frontend/app/utils/locale-utils.ts
```diff
@@ -1,8 +1,9 @@
-import { type FlatNamespace, type Namespace, createInstance } from 'i18next';
+import { type FlatNamespace, createInstance } from 'i18next';
 import I18nextBrowserLanguageDetector from 'i18next-browser-languagedetector';
 import I18NextHttpBackend from 'i18next-http-backend';
 import { initReactI18next } from 'react-i18next';
 
+import { i18nNamespacesSchema } from './route-utils';
```
gregory-j-baker, line 6:
```suggestion
import { i18nNamespacesSchema } from '~/utils/route-utils';
```

frontend/app/utils/locale-utils.ts
```diff
@@ -52,6 +53,7 @@ export async function initI18n(namespaces: Array<string>) {
     .init({
       appendNamespaceToMissingKey: true,
       debug: I18NEXT_DEBUG,
+      defaultNS: 'common',
```
gregory-j-baker, line 56: I feel like this should either be `defaultNS: false` or `defaultNS: 'gcweb'`. Preferrably `false`.

Eventually we will git rid of the `common` namespace and replace it with something that is a little more descriptive.

gregory-j-baker (changes requested):

frontend/app/routes/_gcweb-app.$.tsx
sebastien-comeau: We're using arrow function in a coulpe of places and I fixed them. `request` object is not used in this case and it's why I removed it. I added it as you're suggested.

frontend/app/utils/locale-utils.ts
sebastien-comeau: I set `defaultNS: false`

frontend/app/routes/_gcweb-app.$.tsx
gregory-j-baker: It's not terrible to remove it, but I find it helps when debugging issues to have it available. The eslint rule that barfs when a function parameter is not used is turned off for the project (as it should be, imo) so having the parameter there isn't a problem.

gregory-j-baker (approved):
Well done my friend. Well done.

PR #84, opened 2024-01-17
User service with dependencies

sebastien-comeau:
User service with dependencies to follow the dependency injection convention.


frontend/app/services/user-service.ts
```diff
@@ -0,0 +1,47 @@
...
+  firstName: z.string().optional(),
+  lastName: z.string().optional(),
+  phoneNumber: z.string().optional(),
+});
+
+export type UserInfo = z.infer<typeof userInfoSchema>;
+
+export interface UserServiceDependencies {
+  readonly env: ServerEnv;
+}
```
gregory-j-baker, line 15: Why not use `type`?

---

Also, I don't know if I agree with requiring `ServerEnv` here. The type should contain only what is required for the service to function.

For example, if `ServerEnv` *requires* that some unrelated property be set, then when testing we would have to set that property.

ex: `const { .. } = userService({ displayBetaBanner: false, ...rest })`

gregory-j-baker: <https://www.typescriptlang.org/play>

> // That said, we recommend you use interfaces over type
> // aliases. Specifically, because you will get better error
> // messages. If you hover over the following errors, you can
> // see how TypeScript can provide terser and more focused
> // messages when working with interfaces like Chicken.

This is an official Microsoft documentation site, so I feel like it is authoritative. I'll resolve my comment.

gregory-j-baker: Unresolving because of the second part of my comment (re: `ServerEnv`).

sebastien-comeau: Is this okay?

```
export interface UserServiceDependencies {
  readonly env: Pick<ServerEnv, 'INTEROP_API_BASE_URI'>;
}
```

sebastien-comeau: Fixed

sebastien-comeau: > https://www.typescriptlang.org/play
> 
> > // That said, we recommend you use interfaces over type
> > // aliases. Specifically, because you will get better error
> > // messages. If you hover over the following errors, you can
> > // see how TypeScript can provide terser and more focused
> > // messages when working with interfaces like Chicken.
> 
> This is an official Microsoft documentation site, so I feel like it is authoritative. I'll resolve my comment.

I can't find that quote. 🤷

gregory-j-baker (approved):

PR #88, opened 2024-01-18
Change `headers.set()` to `headers.append()` in `entry.server.tsx`

gregory-j-baker:
Change `headers.set()` to `headers.append()` in `entry.server.tsx` in preparation for upcoming server-side session implementation.

Reasoning: `headers.set()` will overwrite any existing headers of the same name. Since we will need to set multiple `Set-Cookie` headers in the response, we have to use `headers.append()` Since I have to change this for one line of code, I decided to change them all just for consistency.

Related ADO workitems

### DIFF
--- a/frontend/app/entry.server.tsx
+++ b/frontend/app/entry.server.tsx
@@ -45,9 +45,10 @@ export default async function handleRequest(request: Request, responseStatusCode
   responseHeaders.set('Content-Security-Policy', contentSecurityPolicy);
   responseHeaders.set('Content-Type', 'text/html; charset=UTF-8');
   responseHeaders.set('Referrer-Policy', 'strict-origin-when-cross-origin');
-  responseHeaders.set('Set-Cookie', langCookie);
   responseHeaders.set('X-Content-Type-Options', 'nosniff');
   responseHeaders.set('X-Frame-Options', 'deny');
+  // .append() because there can be more than one cookie in a response
+  responseHeaders.append('Set-Cookie', langCookie);
 
   return new Promise((resolve, reject) => {
     let shellRendered = false;


sebastien-comeau:
I will approve. Maybe `set` is ideal over `append` in some circumstances.

https://developer.mozilla.org/en-US/docs/Web/API/Headers/append

> If the specified header already exists, append() will change its value to the specified value. If the specified header already exists and accepts multiple values, append() will append the new value to the end of the value set:

```js
myHeaders.append("Accept-Encoding", "deflate");
myHeaders.append("Accept-Encoding", "gzip");
myHeaders.get("Accept-Encoding"); // Returns 'deflate, gzip'
```

> To overwrite the old value with a new one, use [Headers.set](https://developer.mozilla.org/en-US/docs/Web/API/Headers/set).

sebastien-comeau (dismissed):

gregory-j-baker:
> I will approve. Maybe `set` is ideal over `append` in some circumstances.
> 
> https://developer.mozilla.org/en-US/docs/Web/API/Headers/append
> 
> > If the specified header already exists, append() will change its value to the specified value. If the specified header already exists and accepts multiple values, append() will append the new value to the end of the value set:
> 
> ```js
> myHeaders.append("Accept-Encoding", "deflate");
> myHeaders.append("Accept-Encoding", "gzip");
> myHeaders.get("Accept-Encoding"); // Returns 'deflate, gzip'
> ```
> 
> > To overwrite the old value with a new one, use [Headers.set](https://developer.mozilla.org/en-US/docs/Web/API/Headers/set).

I consider that, but liked the consistency of using `.append()` everywhere. Thinking about it again, you're right. I will change my implementation.

frontend/app/entry.server.tsx
```diff
@@ -45,9 +45,10 @@ export default async function handleRequest(request: Request, responseStatusCode
   responseHeaders.set('Content-Security-Policy', contentSecurityPolicy);
   responseHeaders.set('Content-Type', 'text/html; charset=UTF-8');
   responseHeaders.set('Referrer-Policy', 'strict-origin-when-cross-origin');
-  responseHeaders.set('Set-Cookie', langCookie);
   responseHeaders.set('X-Content-Type-Options', 'nosniff');
   responseHeaders.set('X-Frame-Options', 'deny');
+  // .append() because there can be more than one cookie in a response
+  responseHeaders.append('Set-Cookie', langCookie);
```
sebastien-comeau, line 51: Have you tested the language switching with cookie since we may have more than one `Set-Cookie` for `langCookie`?

sebastien-comeau: Maybe the later `langCookie` set win. 🤷

gregory-j-baker: Remix doesn't clobber existing cookies when you return a `Set-Cookie` header in an action/loader: <https://github.com/remix-run/remix/blob/main/packages/remix-server-runtime/headers.ts#L35>

sebastien-comeau (approved):

PR #89, opened 2024-01-18
Format and rename `server.ts`

gregory-j-baker:
Formatted `server.ts` and then renamed it to `express.server.ts`.


Fbasham (approved):
Builds and runs as intended.

PR #94, opened 2024-01-18
Initial implementation of `redis-service.server.ts`

gregory-j-baker:
This is the initial implementation of a Redis client module. Currently it supports connecting and `get`/`set`/`del` operations.

Future enhancements should include connection pool configuration.

## Related ADO workitems:

### DIFF
--- a/frontend/app/services/redis-service.server.ts
+++ b/frontend/app/services/redis-service.server.ts
@@ -1,0 +1,72 @@
+/**
+ * The RedisService is a service module for interacting with a Redis server. It
+ * provides methods for getting, setting, and deleting values from Redis.
+ *
+ * The getRedisService() function creates a singleton instance of the Redis
+ * client. The instance is lazy initialized as needed, so it will only be
+ * created when a client requests it. This ensures that a single connection pool
+ * is used for the entire application.
+ *
+ * Example usage:
+ *
+ *   const redisService = getRedisService({ url: 'redis://redis.example.com' });
+ *   await redisService.set('key', 'value', { EX: 600 }); // expire in 10 minutes
+ *   const value = await redisService.get('key');
+ *   await redisService.del('key');
+ *
+ * @see https://redis.io/docs/connect/clients/nodejs/
+ */
+import { type SetOptions, createClient } from 'redis';
+
+import { getEnv } from '~/utils/env.server';
+import { getLogger } from '~/utils/logging.server';
+
+const log = getLogger('redis-service.server');
+
+/**
+ * Singleton redis client instance; lazy initialized as needed.
+ */
+let redisClient: ReturnType<typeof createClient> | undefined = undefined;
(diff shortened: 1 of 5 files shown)


frontend/app/services/redis-service.server.ts
```diff
@@ -0,0 +1,71 @@
...
+  password?: string;
+}
+
+export function getRedisService(config: RedisServiceConfigurationOptions) {
+  return {
+    /**
+     * @see https://redis.io/commands/get/
+     */
+    get: async (key: string) => {
+      redisClient ??= await createRedisClient(config.url, config.username, config.password);
```
sebastien-comeau, line 42: Create this inside `getRedisService` and reuse it.

```ts
async function getRedisClient() {
 return redisClient ??= await createRedisClient(config.url, config.username, config.password);
}
```

frontend/app/services/redis-service.server.ts
```diff
@@ -0,0 +1,71 @@
...
+      const value = await redisClient.get(key);
+      return value ? JSON.parse(value) : null;
+    },
+    /**
+     * @see https://redis.io/commands/set/
+     */
+    set: async (key: string, value: unknown, options?: SetOptions) => {
+      redisClient ??= await createRedisClient(config.url, config.username, config.password);
+      return await redisClient.set(key, JSON.stringify(value), options);
```
sebastien-comeau, line 51: You don't need to `await` before a return.

sebastien-comeau (changes requested):

frontend/app/services/redis-service.server.ts
gregory-j-baker: Fixed in latest push

frontend/app/services/redis-service.server.ts
gregory-j-baker: Was left over from a refactoring. I removed it in my latest push.

frontend/app/services/redis-service.server.ts
```diff
@@ -0,0 +1,72 @@
...
+
+import { getEnv } from '~/utils/env.server';
+import { getLogger } from '~/utils/logging.server';
+
+const log = getLogger('redis-service.server');
+
+/**
+ * Singleton redis client instance; lazy initialized as needed.
+ */
+let redisClient: ReturnType<typeof createClient>;
```
sebastien-comeau, line 29:
```suggestion
let redisClient: ReturnType<typeof createClient> | undefined = undefined;
```

sebastien-comeau (changes requested):

frontend/app/services/redis-service.server.ts
gregory-j-baker: Fixed in latest push.

sebastien-comeau (approved):

PR #96, opened 2024-01-18
add loader unit test for personal information page

Fbasham:
Add a unit test for the loader data in the personal information display page.


gregory-j-baker (approved):
Looks okay to me, although I would prefer some spacing between the tests to it's a little easier to skim.

PR #100, opened 2024-01-19
AB#2673 -- session management

gregory-j-baker:
## Initial implementation of application session management.

This commit adds three types of session management to the application: `memory`, `file`, and `redis`.

Configuration for each type of session is handled through environment variables starting with `SESSION_*`, with sane defaults provided for all configurations.

~A basic unit test is included in this PR, but coverage is not yet 100% (need to validate that redis `get`, `set`, and `del` are called when expected).~ **EDIT:** I have force-pushed a new commit with a more exhaustive set of tests.

### Future PR considerations

Sessions must be manually committed, as per [the Remix docs](https://remix.run/docs/en/main/utils/sessions#using-sessions). A future improvement would be to auto-commit sessions in `express.server.ts` by firing the `commitSession(..)` automatically in middleware *after* the Remix request handler has executed.

### Related ADO workitems:

### DIFF
--- a/frontend/app/services/session-service.server.ts
+++ b/frontend/app/services/session-service.server.ts
@@ -1,0 +1,89 @@
+/**
+ * SessionService is a service module responsible for managing user sessions.
+ * Sessions can be stored in memory, as files on disk, or in a Redis database.
+ *
+ * All createSessionStorage() functions take a cookie object, which is used to
+ * create browser's the session cookie. The session cookie is then used to track
+ * the user's session.
+ *
+ * The SessionService module provides functions for creating, reading,
+ * updating, and deleting session data.
+ *
+ * Example usage:
+ *
+ *    const sessionStorage = await getSessionService().createSessionStorage();
+ *    const session = await sessionStorage.getSession(request.headers.get('Cookie'));
+ *    return json({}, {
+ *      headers: {
+ *        'Set-Cookie': await sessionStorage.commitSession(session)
+ *      }
+ *    });
+ *
+ * @see https://remix.run/docs/en/main/utils/sessions
+ */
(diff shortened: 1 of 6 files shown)


frontend/app/services/session-service.server.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  return {
+    createSessionStorage: () => {
+      switch (env.SESSION_STORAGE_TYPE) {
+        case 'memory':
+          // note: memory-backed sessions are not cleaned up; so use only during development!
+          return createMemorySessionStorage({ cookie: sessionCookie });
+        case 'file':
+          // note: memory-backed sessions are not cleaned up; so use only during development!
+          return createFileSessionStorage({ cookie: sessionCookie, dir: env.SESSION_FILE_DIR });
```
nicholas-ly, line 83: **Nitpick:** This comment regarding in-memory sessions can be removed for file sessions

nicholas-ly (dismissed):

frontend/app/services/session-service.server.ts
gregory-j-baker: That comment should be:

```
// note: file-backed sessions are not cleaned up; so use only during development!
```

Copy/paste error. :D (I'll fix)

gregory-j-baker: [Fixed](https://github.com/DTS-STN/canadian-dental-care-plan/pull/100/commits/c1d4a3c63d24c38a0c3d4472f3f5463aec365178#diff-0907994366d0437200280761d2f5478813c8b9e35737595580998eaf4ad7d7c0R81-R85)

nicholas-ly (approved):

PR #103, opened 2024-01-19
AB#2778  --Create landing page with links for the various features of CDCP

olivierroyESDC:
Modified the index file of CDCP to add more links.

### DIFF
--- a/frontend/app/routes/_gcweb-app._index.tsx
+++ b/frontend/app/routes/_gcweb-app._index.tsx
@@ -6,6 +6,8 @@ import { useTranslation } from 'react-i18next';
 import { userService } from '~/services/user-service.server';
 import { getTypedI18nNamespaces } from '~/utils/locale-utils';
 
+import { LandingPageLink } from '~/components/landing-page-link';
+
 const i18nNamespaces = getTypedI18nNamespaces('common', 'gcweb');
 
 export const handle = {
@@ -38,12 +40,30 @@ export default function Index() {
       <p>
         Welcome {userInfo.firstName} {userInfo.lastName}
       </p>
-      <ul>
+      <div className="grid gap-4 md:grid-cols-2">
+        <LandingPageLink title="Update Your Information" description="In this section, you can your personal information." to="/update-info">
+          Update your information
+        </LandingPageLink>
+        <LandingPageLink title="Upload Document" description="In this section, you can your upload documents." to="/Upload-Document">
+          Upload document for CDCP
+        </LandingPageLink>
+        <LandingPageLink title="Personal Information" description="In this section, you can see your personal information." to="/Personal-Information">
+          Personal information
+        </LandingPageLink>
+        <LandingPageLink title="View my Letters" description="In this section, you can see any letters that were sent for you." to="/View-Letters">
+          View My Letters
+        </LandingPageLink>
+        <LandingPageLink title="View my messages" description="In this section, you can see any messages that were sent to you." to="/Messages">
+          View My CDCP Messages
+        </LandingPageLink>
+        <LandingPageLink title="CDCP alerts" description="In this section, you can subscribe to the email alerts for CDCO, Any time you have a new message, you will be alerted." to="/Alert-me">
+          Subscribe to CDCP email alerts
+        </LandingPageLink>
+      </div>
+      <h2>LEGACY LINKS</h2>
+      <ul className="list-unstyled">
         <li>
           <Link to="/personal-information">Personal information</Link>
-        </li>
-        <li>
-          <Link to="/update-info">Update personal info</Link>
         </li>
         <li>
           <Link to="/about">About</Link>
--- a/frontend/app/routes/_gcweb-app._index.tsx
+++ b/frontend/app/routes/_gcweb-app._index.tsx
@@ -44,19 +44,19 @@ export default function Index() {
         <LandingPageLink title="Update Your Information" description="In this section, you can your personal information." to="/update-info">
           Update your information
         </LandingPageLink>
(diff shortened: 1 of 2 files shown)


frontend/app/components/landing-page-link-generator.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+    return (
+        <div class="row panel panel-default">
+            <div class="mrgn-tp-md header-bg">
+                <h2 id="a1-4" class="mrgn-tp-0">{title}</h2>
+                <div class="brdr-bttm-fade"></div>
+            </div>
+            <p class="mrgn-tp-md">{description}</p>
+            <li class="mrgn-bttm-sm xs-block">
+                <a class="btn btn-default btn-block" href={buttonUrl} >
```
nicholas-ly, line 14: FYI use `className` instead of `class`, the latter of which is a reserved JS keyword.

gregory-j-baker (changes requested):

frontend/app/components/landing-page-link-generator.tsx
```diff
@@ -0,0 +1,20 @@
+
+
+
+export function GenerateLandingPageLink({ title, description, buttonUrl, buttonText }) {
+
+    return (
+        <div class="row panel panel-default">
+            <div class="mrgn-tp-md header-bg">
+                <h2 id="a1-4" class="mrgn-tp-0">{title}</h2>
```
nicholas-ly, line 9: Good idea separating the landing page links into a component.

Just noticed that if we re-use this component, we will have a duplicate `id` in the `h2`, which will be flagged as an accessibility issue. Consider having some sort of way (a prop perhaps?) to uniquely specify an `id`.

frontend/app/routes/_gcweb-app._index.tsx
```diff
@@ -42,13 +45,16 @@ export default function Index() {
       <p>
         Welcome {userInfo.firstName} {userInfo.lastName}
       </p>
-      <ul>
+      <ul class="list-unstyled list-inline">
+
+        {LANDING_PAGE_LINKS.map((linkItem) => (
+          <GenerateLandingPageLink key={linkItem.title} {...linkItem} />
+        ))}
```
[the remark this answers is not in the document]
olivierroyESDC: I've applied required changes, please review it, thanks

frontend/app/components/landing-page-link-generator.tsx
```diff
@@ -0,0 +1,20 @@
+
+
+
+export function GenerateLandingPageLink({ title, description, buttonUrl, buttonText }) {
```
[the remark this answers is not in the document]
olivierroyESDC: I've applied required changes, please review it, thanks

frontend/app/data/landing-page-links.js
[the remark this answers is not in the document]
olivierroyESDC: file deleted

frontend/app/components/landing-page-link-generator.tsx
olivierroyESDC: I've applied required changes, please review it, thanks

frontend/app/components/landing-page-link-generator.tsx
olivierroyESDC: I've applied required changes, please review it, thanks

frontend/app/components/landing-page-link-generator.tsx
```diff
@@ -0,0 +1,31 @@
...
+                <div className="brdr-bttm-fade"></div>
+            </div>
+            <p className="mrgn-tp-md">{landingPageLinkObject.description}</p>
+            <Link className="btn btn-default btn-block" to={landingPageLinkObject.buttonUrl}>
+                {landingPageLinkObject.buttonText}
+            </Link>
+        </div>
+
+    );
+}
```
sebastien-comeau, line 31: There is many suggestions in this comment.

- Remove the `landingPageLinkObject` in `LandingPageLinkProps` interface
- Extends from `Link` props but omit `children` since we want to make it required. 
- Uses `clsx` package to construct link's `className` strings if it is externally set.
- Removes `buttonUrl` prop in favor of Link's `to`.
- Removes `buttonText` prop in favor of Link's `children`.
- Renames `landing-page-link-generator.tsx` file to `landing-page-link.tsx`

```suggestion
import { type ReactNode } from 'react';

import { Link } from '@remix-run/react';

import clsx from 'clsx';

export interface LandingPageLinkProps extends Omit<React.ComponentProps<typeof Link>, 'children'> {
  children: ReactNode;
  description: string;
  title: string;
}

export function LandingPageLink(props: LandingPageLinkProps) {
  const { className, children, description, title, ...linkProps } = props;
  const linkClassName = clsx('btn btn-default btn-block', className);
  return (
    <div className="row panel panel-default">
      <div className="mrgn-tp-md header-bg">
        <h2 className="mrgn-tp-0">{title}</h2>
        <div className="brdr-bttm-fade"></div>
      </div>
      <p className="mrgn-tp-md">{description}</p>
      <Link className={linkClassName} {...linkProps}>
        {children}
      </Link>
    </div>
  );
}
```

sebastien-comeau (dismissed):

gregory-j-baker:
@olivierroyESDC I'm not sure what's going on with your commit graph, but at the time of this comment, your PR has 54 commits (many of which are already in the main branch).

My suggestion to fix this right now would be to do this:

```
git fetch
git checkout create_landing_page
git reset origin/main # this will NOT destroy your changes
```

This will reset your branch back to what's in main.. then you can stage all of your modified files and commit them in a single commit.

gregory-j-baker (approved):

PR #105, opened 2024-01-19
modify playwright webserver command

Fbasham:
Running into some issues with running e2e test on Windows.  

Running `npm run test:e2e` on windows produces:

```javascript
[WebServer]  info  building... (NODE_ENV=production)
[WebServer]  info  built (4s)
[WebServer] 'MOCKS_ENABLED' is not recognized as an internal or external command,
operable program or batch file.Error: Process from config.webServer was not able to start. Exit code: 1
```

cross-env is needed here to make it play nice with Windows.


gregory-j-baker (approved):
That was my doing. Sorry, I keep forgetting about the windows users (or do I? 😈).

PR #111, opened 2024-01-20
Change root font-size

sebastien-comeau:
Opt for a root font size of `16px` since Tailwind sizes predominantly use rem units. This becomes problematic when the font size is set to `10px`. It's important to mention that using rem units offers better scalability and responsiveness, as it is relative to the root font size. This ensures a more consistent and adaptable design, especially across different devices and screen sizes. Additionally, relying on rem units promotes better accessibility and a more user-friendly experience.


gregory-j-baker (changes requested):
I approve the change, however I would prefer if you applied these styles in a `gcweb-fixes.css` file for maintainability purposes.

If we ever update to a newer version of GCWeb, we don't want to have to redo these changes again.

sebastien-comeau:
> I approve the change, however I would prefer if you applied these styles in a `gcweb-fixes.css` file for maintainability purposes.
> 
> If we ever update to a newer version of GCWeb, we don't want to have to redo these changes again.

Not a bad idea. Will do.

PR #112, opened 2024-01-20
Change root font-size

sebastien-comeau:
Opt for a root font size of 16px since Tailwind sizes predominantly use rem units. This becomes problematic when the font size is set to 10px. It's important to mention that using rem units offers better scalability and responsiveness, as it is relative to the root font size. This ensures a more consistent and adaptable design, especially across different devices and screen sizes. Additionally, relying on rem units promotes better accessibility and a more user-friendly experience.

### DIFF
--- a/frontend/app/cdcp.css
+++ b/frontend/app/cdcp.css
@@ -1,0 +1,3 @@
+html {
+    font-size: 16px;
+}
--- a/frontend/app/routes/_gcweb-app.tsx
+++ b/frontend/app/routes/_gcweb-app.tsx
@@ -5,6 +5,7 @@ import { Link, Outlet, isRouteErrorResponse, useRouteError } from '@remix-run/re
 
 import { Trans, useTranslation } from 'react-i18next';
 
+import cdcpStylesheet from '~/cdcp.css';
 import { LanguageSwitcher } from '~/components/language-switcher';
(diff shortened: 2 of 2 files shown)


frontend/package.json
```diff
@@ -11,6 +11,7 @@
     "dev:server": "cross-env NODE_ENV=development npm run-script build:server -- --watch",
     "start": "cross-env NODE_ENV=production node ./build/express.server.js ./build/index.js",
     "format": "prettier --write ./",
+    "minify-css": "cleancss --batch --batch-suffix \".min\" ./public/theme/cdcp/css/*.css !./public/theme/cdcp/css/*.min.css",
```
gregory-j-baker, line 14: Is this really needed for a one-line css file?

gregory-j-baker (changes requested):

frontend/package.json
sebastien-comeau: It's one line for now... who knows later?

gregory-j-baker: It's never gonna be big enough that we need to care about minification..

sebastien-comeau: - Removed clean-css.
- Created and imported `cdcp.css` stylesheet.

gregory-j-baker (approved):

PR #116, opened 2024-01-21
Refactor user service

gregory-j-baker:
Refactor user service to use ESM import singleton pattern.


sebastien-comeau (approved):
Ok for now but singleton service could/will be a problem when authentication will be implemented.

PR #117, opened 2024-01-21
Refactor redis service

gregory-j-baker:
Refactor redis service to use ESM import singleton pattern.

### DIFF
--- a/frontend/app/services/redis-service.server.ts
+++ b/frontend/app/services/redis-service.server.ts
@@ -21,28 +21,21 @@ import { type SetOptions, createClient } from 'redis';
 import { getEnv } from '~/utils/env.server';
 import { getLogger } from '~/utils/logging.server';
 
-const log = getLogger('redis-service.server');
-
-/**
- * Singleton redis client instance; lazy initialized as needed.
- */
-let redisClient: ReturnType<typeof createClient> | undefined = undefined;
-
-export function getRedisService() {
+function createRedisService() {
+  const log = getLogger('redis-service.server');
   const { REDIS_URL, REDIS_USERNAME, REDIS_PASSWORD } = getEnv();
 
-  function createRedisClient(url: string, username?: string, password?: string) {
-    log.info(`Creating new Redis client; url=[${url}]`);
-    return createClient({ url, username, password })
-      .on('connect', () => log.info(`Redis client initiating connection to [${url}]`))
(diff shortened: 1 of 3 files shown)


frontend/app/services/redis-service.server.ts
```diff
@@ -21,52 +21,47 @@ import { type SetOptions, createClient } from 'redis';
...
-    return (redisClient ??= await createRedisClient(REDIS_URL, REDIS_USERNAME, REDIS_PASSWORD));
-  }
-
   return {
     /**
      * @see https://redis.io/commands/get/
      */
     get: async (key: string) => {
-      const redisClient = await getRedisClient();
+      redisClient ??= await createRedisClient();
```
sebastien-comeau, line 46: I still thing the following function should be call instead of repeating this `redisClient ??= await createRedisClient();` each time you want to use the client.

Don't repeat yourself.
```
async function getRedisClient() {
    return (redisClient ??= await createRedisClient());
  }
```

or do the create and lazy loading in `getRedisClient()`.

sebastien-comeau (changes requested):

frontend/app/services/redis-service.server.ts
gregory-j-baker: The logic around lazy loading is tricky. If you put it in the wrong place, it can accidentally init when you don't expect it to.

I'll move the init into `getRedisClient()`.

gregory-j-baker: Fixed in latest push.

frontend/app/services/redis-service.server.ts
```diff
@@ -59,14 +52,16 @@ export function getRedisService() {
      */
     set: async (key: string, value: unknown, options?: SetOptions) => {
       const redisClient = await getRedisClient();
-      return redisClient.set(key, JSON.stringify(value), options);
+      return await redisClient.set(key, JSON.stringify(value), options);
```
sebastien-comeau, line 55: No need to `await` at the return.

sebastien-comeau (changes requested):

frontend/app/services/redis-service.server.ts
gregory-j-baker: Weird that vscode only warns me about this sometimes.

sebastien-comeau (approved):

PR #118, opened 2024-01-21
test: add input-label component unit test

sebastien-comeau:
As per title.

Addtional note:

- Added [Vitest UI](https://vitest.dev/guide/ui.html#vitest-ui)

### DIFF
--- a/frontend/__tests__/components/input-label.test.tsx
+++ b/frontend/__tests__/components/input-label.test.tsx
@@ -1,5 +1,7 @@
 import { getByTestId, render, screen } from '@testing-library/react';
 
+import { afterEach, describe, expect, it } from 'vitest';
+
 //import { axe, toHaveNoViolations } from 'jest-axe';
 import { InputLabel } from '~/components/input-label';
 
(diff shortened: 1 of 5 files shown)


frontend/__tests__/components/input-label.test.tsx
gregory-j-baker, line 1: Import `describe`, `it`, `expect`, etc from vitest, otherwise you're using jest.

``` typescript
import { describe, expect, it } from 'vitest';
```

frontend/app/components/input-label.tsx
```diff
@@ -16,9 +16,13 @@ export const InputLabel = (props: InputLabelProps) => {
   return (
     <label className={clsx(required && 'required', className)} data-testid="input-label" {...restLabelProps}>
       <span className="field-name">{children}</span>
-      {required && <strong className="required">&#32;({t('input-label.required')})</strong>}
+      {required && (
+        <strong className="required" data-testid="input-label-required">
+          &#32;({t('input-label.required')})
```
gregory-j-baker, line 21: Why not just put a small left margin on the element to emulate a space?

gregory-j-baker (changes requested):

frontend/app/components/input-label.tsx
gregory-j-baker: Marking as `/wontfix` ... we can revisit later.

gregory-j-baker (approved):

frontend/app/components/input-label.tsx
sebastien-comeau: Without it the label text will be read like "Email address(Required)". 🤷

PR #121, opened 2024-01-22
Minor typography fix

gregory-j-baker:
See: <https://wet-boew.github.io/wet-boew-styleguide/design/forms-en.html#:~:text=Place%20%22(required)%22%20at%20the%20end%20of%20the%20field%20label%20if%20information%20is%20required.>


sebastien-comeau (approved):
Good catch. I was looking at the server-side form validation example.

https://wet-boew.github.io/v4.0-ci/demos/formvalid/formvalid-server-en.html

image

PR #124, opened 2024-01-22
Increase coverage for language-switcher.test.tsx

gregory-j-baker:
Previous iteration wasn't testing search param handling.


Fbasham (approved):
Sweet.  Learned something new about mocks.

PR #125, opened 2024-01-22
AB#2769 add preferred language view

Fbasham:
This PR creates a new route:  `/personal-information/preferred-language`, and displays the user's preferred language.  There isn't much to render (yet), as designs haven't been finalized, so this PR is intended as more of a scaffold for future iterations.

## Changes
- add `/personal-information/preferred-language` route
- add new locale(s) files (`personal-information.json`)
- add new locale type to type declarations
- stub out e2e test for a11y checking

### DIFF
--- a/frontend/app/routes/_gcweb-app.personal-information.preferred-language.tsx
+++ b/frontend/app/routes/_gcweb-app.personal-information.preferred-language.tsx
@@ -1,0 +1,44 @@
+import { type LoaderFunctionArgs, json } from '@remix-run/node';
+import { useLoaderData } from '@remix-run/react';
+
+import { useTranslation } from 'react-i18next';
+
+import { userService } from '~/services/user-service.server';
+import { getTypedI18nNamespaces } from '~/utils/locale-utils';
+
+const i18nNamespaces = getTypedI18nNamespaces('common', 'personal-information');
+
+export const handle = {
+  breadcrumbs: [{ labelI18nKey: 'common:personal-information.breadcrumbs.home', to: '/' }, { labelI18nKey: 'common:personal-information.page-title', to: '/personal-information' }, { labelI18nKey: 'common:preferred-language.page-title' }],
+  i18nNamespaces,
+  pageIdentifier: 'CDCP-0004',
+  pageTitleI18nKey: 'common:preferred-language.page-title',
+} as const satisfies RouteHandleData;
+
+export async function loader({ request }: LoaderFunctionArgs) {
+  const userId = await userService.getUserId();
+  const userInfo = await userService.getUserInfo(userId);
+
+  if (userInfo === null) {
+    throw new Response(null, { status: 404 });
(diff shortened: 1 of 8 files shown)


nicholas-ly (dismissed):
Looks good!
/approved

My only comment would require agreement from the team:
Should we consider using a shorter name (ie. `/language` instead of /`preferred-language` and down the line `/documents` instead of `/upload-documents`)?

gregory-j-baker:
> Should we consider using a shorter name

IMO it doesn't matter at all since ain't nobody gonna be typing in the URL manually.

frontend/app/routes/_gcweb-app.personal-information.preferred-language.tsx
```diff
@@ -0,0 +1,39 @@
...
+  breadcrumbs: [{ labelI18nKey: 'common:personal-information.breadcrumbs.home', to: '/' }, { labelI18nKey: 'common:personal-information.page-title', to: '/personal-information' }, { labelI18nKey: 'common:preferred-language.page-title' }],
+  i18nNamespaces,
+  pageIdentifier: 'CDCP-0004',
+  pageTitleI18nKey: 'common:preferred-language.page-title',
+} as const satisfies RouteHandleData;
+
+export async function loader({ request }: LoaderFunctionArgs) {
+  const userId = await userService.getUserId();
+  const userInfo = await userService.getUserInfo(userId);
+  return json({ user: userInfo });
```
sebastien-comeau, line 21: Should throw `404` response if `userInfo` is not found.

sebastien-comeau (changes requested):

gregory-j-baker (dismissed):

frontend/app/routes/_gcweb-app.personal-information.preferred-language.tsx
gregory-j-baker: Not wrong, but the entire application will require user data.. if we don't have it, we should probably catch that and handle it in the root route by maybe displaying a "you are not eligible to use this service" page.

sebastien-comeau: Yup but we are not there yet. Everything could/would go to garbage at any time. 🤷‍♂️

sebastien-comeau: @Fbasham remove this optional chaining (?.) since user shouldn't be `null` or `undefined`. :)

```tsx
<dd>{user?.preferredLanguage}</dd>
```

sebastien-comeau (approved):

frontend/app/routes/_gcweb-app.personal-information.preferred-language.tsx
Fbasham: kk, added.  Anything else?

PR #131, opened 2024-01-22
Translaton on every page.

nibivi77:
[ADO Link](https://dev.azure.com/DTS-STN/Canada%20Dental%20Care%20Plan/_workitems/edit/2805)

[AB#2805](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/2805)

Added translation on all pages.

Notes
 - Created two more locale files to follow route naming convention. (Open to suggestions on this.)
 - Tests will follow in a separate PR.
 - `phone-number.tsx` component has translation in it. Should the text only be passed down as props from the pages to the component?


frontend/app/components/phone-number.tsx
```diff
@@ -6,25 +11,26 @@ export type PhoneNumberProps = {
...
-            <span className="field-name">Previous phone number</span>
+            <span className="field-name">{t('update-phone-number:component.previous')}</span>
           </label>
           <input id="previousPhoneNumber" name="previousPhoneNumber" className="form-control mrgn-bttm-sm" maxLength={32} defaultValue={previousPhoneNumber} data-testid="previousPhoneNumber" disabled={true} />
         </>
       )}
       <label htmlFor="phoneNumber" className={editMode ? 'required' : 'none'}>
-        <span className="field-name">{previousPhoneNumber !== undefined ? 'New phone number' : 'Phone number'}</span>
-        {editMode && <strong className="required mrgn-lft-sm">(required)</strong>}
+        <span className="field-name">{previousPhoneNumber !== undefined ? `${t('update-phone-number:component.previous')}` : `${t('update-phone-number:component.phone')}`}</span>
```
gregory-j-baker, line 26:
```suggestion
        <span className="field-name">
          {t(previousPhoneNumber !== undefined ? 'update-phone-number:component.previous' : 'update-phone-number:component.phone')}
        </span>

```

frontend/app/routes/_gcweb-app._index.tsx
```diff
@@ -36,23 +36,23 @@ export default function Index() {
         {t('common:index.page-title')}
       </h1>
       <p>
-        Welcome {userInfo.firstName} {userInfo.lastName}
+      {t('common:index.welcome')} {userInfo.firstName} {userInfo.lastName}
```
gregory-j-baker, line 39: This page should only use translations from its own namespace, and *maybe* the `gcweb` one (for common error messages).

Every page's translations should be isolated, even if it means repeating some strings in different .json files. The reason is that if we have a `common.json` file that contains a string, and that string is used by multiple pages, when we update that string all of those pages are affected.

gregory-j-baker, line 39: Translations that include data should have the data inside of the translation.. like this:

`"my-string": "Welcome <firstName>John</firstName> <lastName>Doe</lastName>"`

and then the data should be substituted into the string using a `<Trans>` component.

The reason is that english and french have different grammar rules. The example I always use is this:

```
en: "The red <thing>object</thing>"
fr: "Le <thing>object</thing> route"
```

frontend/app/routes/_gcweb-app.personal-information.address.confirm.tsx
```diff
@@ -23,20 +27,20 @@ export async function action({ request }: ActionFunctionArgs) {
...
-
+  const { t } = useTranslation(i18nNamespaces);
   return (
     <>
       <h1 id="wb-cont" property="name">
-        Change address
+      {t('personal-information:edit.change-address')}
       </h1>
-      <h2 className="h3">Confirm</h2>
+      <h2 className="h3">{t('gcweb:input-label.button.confirm')}</h2>
```
gregory-j-baker, line 36: As alluded to in a previous comment, even though `Confirm` might be a common button label across the application, we should still define it again in this view's own translation file.

Years ago I worked on an application where they wanted to change the label of a cancel button to "go back" and the developer made the change and it altered 15+ pages in the application (and they had to roll back).

frontend/app/routes/_gcweb-app.personal-information.address.edit.tsx
```diff
@@ -61,21 +65,21 @@ export default function ChangeAddress() {
   return (
     <>
       <h1 id="wb-cont" property="name">
-        Change address
+        {t('personal-information:edit.change-address')}
       </h1>
       <Form method="post">
-        <InputField id="home-address" label="Home address" name="homeAddress" className="!w-full lg:!w-1/2" required defaultValue={defaultValues.homeAddress} errorMessage={errorMessages.homeAddress} />
-        <InputField id="mailing-address" label="Mailing address" name="mailingAddress" className="!w-full lg:!w-1/2" required defaultValue={defaultValues.mailingAddress} errorMessage={errorMessages.mailingAddress} />
+        <InputField id="home-address" label={`${t('personal-information:index.home-address')}`} name="homeAddress" className="!w-full lg:!w-1/2" required defaultValue={defaultValues.homeAddress} errorMessage={errorMessages.homeAddress} />
```
gregory-j-baker, line 71: Here, you can just do this:

```
label={t('personal-information:index.home-address')}
```

frontend/app/routes/_gcweb-app.personal-information.address.edit.tsx
```diff
@@ -61,21 +65,21 @@ export default function ChangeAddress() {
...
+        <InputField id="mailing-address" label={`${t('personal-information:index.mailing-address')}`} name="mailingAddress" className="!w-full lg:!w-1/2" required defaultValue={defaultValues.mailingAddress} errorMessage={errorMessages.mailingAddress} />
```
gregory-j-baker, line 72:
```suggestion
        <InputField id="mailing-address" label={t('personal-information:index.mailing-address')} name="mailingAddress" className="!w-full lg:!w-1/2" required defaultValue={defaultValues.mailingAddress} errorMessage={errorMessages.mailingAddress} />
```

frontend/public/locales/en/common.json
gregory-j-baker, line 1: A lot of what is in here should be moved out into view-specific .json files. I kinda regret naming this `common.json`

frontend/public/locales/en/gcweb.json
```diff
@@ -47,6 +47,14 @@
...
-    "required": "required"
+    "error": "Error",
+    "required": "required",
+    "button": {
+      "submit": "Submit",
+      "cancel": "Cancel",
+      "change": "Change",
+      "confirm": "Confirm",
+      "edit": "Edit"
+    }
```
gregory-j-baker, line 58: I don't believe these belong in this file.

gregory-j-baker (changes requested):
Overall a great effort. Some of the changes I suggested can be made in a future PR, if you want. We're in the early days of development so I think it's fine to push some commits through and fix them later.

Also: consider enabling auto-formatting-on-save in your IDE. As much as I dislike that feature, it's what we've decided on for the project.

frontend/app/routes/_gcweb-app._index.tsx
sebastien-comeau: @gregory-j-baker, only use `Trans` from `react-i18next` if you need to render html which is not the case here. The basic `i18next` interpolation should be used here.

https://www.i18next.com/translation-function/interpolation#basic

Keys, by default, are strings surrounded by curly brackets:
```json
"welcome": "Welcome {{firstName}} {{lastName}}",
```

`t` function call:
```ts
t('common:index.welcome', { firstName: userInfo.firstName, lastName: userInfo.firstName });
```

Trans Component
https://react.i18next.com/latest/trans-component

> **Important note**
> While <Trans> gives you a lot of power by letting you interpolate or translate complex React elements, the truth is: in most cases you don't even need it.

frontend/app/routes/_gcweb-app.personal-information.address.confirm.tsx
sebastien-comeau: I agree with @gregory-j-baker. From my experience the confirm or action text should be specific for his context.

frontend/app/routes/_gcweb-app._index.tsx
gregory-j-baker: Ahh yes. I actually completely forgot that you could do this with the `t(..)` function. Thanks for the tips.

frontend/public/locales/en/gcweb.json
sebastien-comeau: This key is specific for the `InputLabel` GCWeb's component that I've created. No more keys should be added to `"input-label"`.

```json
"input-label": {
  "required": "required"
 }
```

frontend/public/locales/en/index.json
```diff
@@ -0,0 +1,23 @@
...
+  "personal-info": "Personal information",
+  "personal-info-desc": "In this section, you can see your personal information.",
+  "update-info": "Update personal information",
+  "update-info-desc": "In this section, you can your personal information.",
+  "upload": "Upload document for CDCP",
+  "upload-desc": "In this section, you can your upload documents.",
+  "view-letters": "View my letters",
+  "view-letters-desc": "In this section, you can see any letters that were sent for you.",
+  "view-CDCP": "View my CDCP messages",
+  "view-CDCP-desc": "In this section, you can see any messages that were sent to you.",
```
gregory-j-baker, line 16: **Nitpick:** the convention is to use kebab-case for these keys. In this case, it would be `view-cdcp` and `view-cdcp-desc`.

gregory-j-baker (approved):
Approved with one nitpick.. I'll insta-resolve it and we can fix it in a future PR.

frontend/public/locales/en/index.json
gregory-j-baker: Insta-resolving this so the PR can be merged. We can change the i18n key in a future PR.

nibivi77: I'll make the change while resolving conflicts.

nibivi77:
`phone-number.tsx` has been removed since this PR was created. I will re-work the naming convention in a future PR.

gregory-j-baker:
Going forward, rebase your branch on top of main.. Merging main into your branch causes all kinds of weird stuff with the diffs and also sometimes re-signs other people's commits with your own private key. I'll approve this time though.

image

gregory-j-baker (approved):

PR #133, opened 2024-01-23
AB#2800, AB#2824 Adding Lookup services with GET lookups/preferred-language

carolinegeorge-gov:
[AB#2800](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/2800)- Added the GET /lookups/preferred-languages mocks
[AB#2824](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/2824) - Added the lookup service

### DIFF
--- a/frontend/app/services/lookup-service.server.ts
+++ b/frontend/app/services/lookup-service.server.ts
@@ -1,10 +1,9 @@
 import { z } from 'zod';
 
 import { getLogger } from '~/utils/logging.server';
+import { getEnv } from '~/utils/env.server';
 
-const logger = getLogger('user-service.server');
-
-export const preferredLanguageSchema = z.object({
+const preferredLanguageSchema = z.object({
   id: z.string().optional(),
   nameEn: z.string().optional(),
   nameFr: z.string().optional(),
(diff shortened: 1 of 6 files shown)


frontend/app/services/lookup-service.server.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export function getLookupService({ env }: LookupServiceDependencies) {
+  const { INTEROP_API_BASE_URI } = env;
+
+  async function getAllPreferredLanguages() {
+    const url = `${INTEROP_API_BASE_URI}/lookups/preferred-languages/`;
+    const response = await fetch(url);
+    const body = await response.json();
+    const preferredLanguageSchemaList = z.array(preferredLanguageSchema);
+    if (response.ok) return {initalValues: preferredLanguageSchemaList.parse(body) };
```
sebastien-comeau, line 29: Get, parse and return parsed JSON if response is OK.
```suggestion
    if (response.ok) return preferredLanguageSchemaList.parse(await response.json());
```

frontend/app/services/lookup-service.server.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export type PreferredLanguageInfo = z.infer<typeof preferredLanguageSchema>;
+
+export interface LookupServiceDependencies {
+  readonly env: {
+    readonly INTEROP_API_BASE_URI: string;
+  };
+}
+
+export function getLookupService({ env }: LookupServiceDependencies) {
```
sebastien-comeau, line 21: @gregory-j-baker changed the service pattern and don't requires to pass service dependencies as argument. He also return the service as singleton at the end of the file. Update this file to match this new pattern.

https://github.com/DTS-STN/canadian-dental-care-plan/blob/main/frontend/app/services/user-service.server.ts

sebastien-comeau (changes requested):

frontend/app/services/lookup-service.server.ts
carolinegeorge-gov: Fixed: 72170e807b3f47c361f34e4bd4d9036444475779

frontend/app/services/lookup-service.server.ts
carolinegeorge-gov: Fixed: 72170e807b3f47c361f34e4bd4d9036444475779

sebastien-comeau (approved):
Good job!

PR #145, opened 2024-01-24
tie up loose ends between view and edit pages

Fbasham:
/This PR is the result of a comment in slack requesting the links between views and edit pages:

image

## Changes
- added views for `/personal-information/address`, `/personal-information/phone-number`
- linked to the respective 'edit' pages for both views

### Additional information 
- page identifiers will definitely need to be changed (I'm not doing it here since there's no point until that convention is finalized)
- locale structuring will definitely need to be refined and cleaned in a future PR across all pages

### DIFF
--- a/frontend/app/routes/_gcweb-app.personal-information._index.tsx
+++ b/frontend/app/routes/_gcweb-app.personal-information._index.tsx
@@ -19,6 +19,10 @@ export async function loader({ request }: LoaderFunctionArgs) {
   const userId = await userService.getUserId();
   const userInfo = await userService.getUserInfo(userId);
 
+  if (!userInfo) {
+    throw new Response(null, { status: 404 });
+  }
+
   return json({ user: userInfo });
 }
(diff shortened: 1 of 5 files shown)


frontend/app/routes/_gcweb-app.personal-information._index.tsx
sebastien-comeau, line 1: Add this following check in the loader and throw 404 reponse if user info is not found. I will remove uncessary optional chaining (?.) in the component.
```tsx
if (!userInfo) {
    throw new Response(null, { status: 404 });
  }
```

sebastien-comeau (changes requested):

frontend/app/routes/_gcweb-app.personal-information._index.tsx
Fbasham: change made

sebastien-comeau (approved):

PR #155, opened 2024-01-24
Only log creation of Redis client when it is created

nicholas-ly:
Every time `getRedisClient()` was called,  "Creating new Redis client..." was logged even if a Redis client already exists.

### DIFF
--- a/frontend/app/services/redis-service.server.ts
+++ b/frontend/app/services/redis-service.server.ts
@@ -28,14 +28,18 @@ function createRedisService() {
   // singleton redis client instance; lazy initialized as needed
   let redisClient: ReturnType<typeof createClient> | undefined = undefined;
 
-  async function getRedisClient() {
+  async function createRedisClient() {
     log.info(`Creating new Redis client; url=[${REDIS_URL}]`);
-    return (redisClient ??= await createClient({ url: REDIS_URL, username: REDIS_USERNAME, password: REDIS_PASSWORD })
+    return createClient({ url: REDIS_URL, username: REDIS_USERNAME, password: REDIS_PASSWORD })
       .on('connect', () => log.info(`Redis client initiating connection to [${REDIS_URL}]`))
       .on('ready', () => log.info('Redis client is ready to use'))
       .on('reconnecting', () => log.info(`Redis client is reconnecting to [${REDIS_URL}]`))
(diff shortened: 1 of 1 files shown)


gregory-j-baker:
This is leftover from a refactoring that I did.. I do think it's important to log when we are creating a new singleton, so maybe we should do something like this:

``` typescript
async function createRedisClient() {
  // log "creating client"
  return createClient({ url: REDIS_URL, username: REDIS_USERNAME, password: REDIS_PASSWORD })
}

async function initRedisClient() {
  redisClient ??= await createRedisClient()
  // rest of the function
}
```

nicholas-ly:
> This is leftover from a refactoring that I did.. I do think it's important to log when we are creating a new singleton, so maybe we should do something like this:
> 
> ```ts
> async function createRedisClient() {
>   // log "creating client"
>   return createClient({ url: REDIS_URL, username: REDIS_USERNAME, password: REDIS_PASSWORD })
> }
> 
> async function initRedisClient() {
>   redisClient ??= await createRedisClient()
>   // rest of the function
> }
> ```

Good idea. Fixed.

frontend/app/services/redis-service.server.ts
```diff
@@ -28,14 +28,18 @@ function createRedisService() {
   // singleton redis client instance; lazy initialized as needed
   let redisClient: ReturnType<typeof createClient> | undefined = undefined;
 
-  async function getRedisClient() {
+  async function createRedisClient() {
     log.info(`Creating new Redis client; url=[${REDIS_URL}]`);
-    return (redisClient ??= await createClient({ url: REDIS_URL, username: REDIS_USERNAME, password: REDIS_PASSWORD })
+    return await createClient({ url: REDIS_URL, username: REDIS_USERNAME, password: REDIS_PASSWORD })
```
gregory-j-baker, line 33: No need to `await` if your wrapping function is `async`.. just return the promise.

gregory-j-baker (changes requested):
Approved, but you can do better!

gregory-j-baker (approved):

frontend/app/services/redis-service.server.ts
nicholas-ly: Noted - thanks

nicholas-ly: And fixed

PR #160, opened 2024-01-25
Enable vitest coverage

sebastien-comeau:


gregory-j-baker (approved):
Approving but it's flaky.. complains about missing sourcemaps if you don't build first.

PR #177, opened 2024-01-26
Switch to ESM build format

gregory-j-baker:
As per title.


Fbasham (approved):
approving with the understanding that this solution works, but it not ideal.  Unfortunately other attempts to get this working are also not ideal.  Will have to revisit in the future if there's time

PR #185, opened 2024-01-26
AB#2827 -- Adding unit tests for loader and action for confirm address route

nicholas-ly:
Note that the actual implementation of `loader(..)` and `action(..)` don't currently check for valid session variables (see https://remix.run/docs/en/main/utils/sessions#sessionhaskey) so these are the most basic unit tests.


Fbasham (approved):
great minds think alike XD

PR #191, opened 2024-01-29
Add attr. type submit to form buttons

sebastien-comeau:
As per title.


sebastien-comeau:
TIL, thanks @gregory-j-baker 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#type

> `submit`: The button submits the form data to the server. This is the default if the attribute is not specified for buttons associated with a `<form>`, or if the attribute is an empty or invalid value.

PR #194, opened 2024-01-29
unify buttons and groupings

Fbasham:
Code refactoring.  Nothing interesting.

### DIFF
--- a/frontend/app/routes/_gcweb-app.personal-information.address.confirm.tsx
+++ b/frontend/app/routes/_gcweb-app.personal-information.address.confirm.tsx
@@ -96,19 +96,13 @@ export default function ConfirmAddress() {
           </div>
         </div>
 
-        <div className="form-group">
-          <ul className="list-inline lst-spcd">
-            <li>
-              <button id="confirm-button" className="btn btn-primary btn-lg">
-                {t('personal-information:address.confirm.button.confirm')}
-              </button>
-            </li>
-            <li>
-              <Link id="cancel-button" to="/personal-information/address/edit" className="btn btn-default btn-lg">
-                {t('personal-information:address.confirm.button.cancel')}
-              </Link>
-            </li>
-          </ul>
+        <div className="flex flex-wrap gap-3">
+          <button id="confirm-button" className="btn btn-primary btn-lg">
+            {t('personal-information:address.confirm.button.confirm')}
+          </button>
+          <Link id="cancel-button" to="/personal-information/address/edit" className="btn btn-default btn-lg">
+            {t('personal-information:address.confirm.button.cancel')}
+          </Link>
         </div>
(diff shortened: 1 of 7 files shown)


frontend/app/routes/_gcweb-app.personal-information.address.confirm.tsx
```diff
@@ -96,19 +96,13 @@ export default function ConfirmAddress() {
...
-              </Link>
-            </li>
-          </ul>
+        <div className="flex flex-wrap gap-3">
+          <button id="confirm-button" className="btn btn-primary btn-lg">
+            {t('personal-information:address.confirm.button.confirm')}
+          </button>
+          <Link id="cancel-button" to="/personal-information/address/edit" className="btn btn-default btn-lg">
+            {t('personal-information:address.confirm.button.cancel')}
+          </Link>
```
nicholas-ly, line 105: FYI those `<ul>` were originally taken from WET's style guide: 
https://wet-boew.github.io/wet-boew-styleguide/design/buttons-en.html#btn-groups

I was under the impression we would be using WET/GCWeb's styles wherever we can and Tailwind when we need to do something that the WET/GCWeb styles do not provide.

CC @sebastien-comeau

nicholas-ly: Resolving since this is a non-blocking comment that may or may not need further discussion.

sebastien-comeau: Using `ul` to align buttons doesn't make sense, IMO. I do agree we should use WET/GCWeb where it make sense but that template is overkill sometimes and it's based on outdated Bootstrap 3.x.

Fbasham: ;) it probably will need further discussion.  @sebastien-comeau, what do you want to do?

nicholas-ly: > Using `ul` to align buttons doesn't make sense, IMO. I do agree we should use WET/GCWeb where it make sense but that template is overkill sometimes and it's based on outdated Bootstrap 3.x.

I'm in full agreement with that.

sebastien-comeau (approved):

frontend/app/routes/_gcweb-app.personal-information.address.confirm.tsx
sebastien-comeau: > ;) it probably will need further discussion. @sebastien-comeau, what do you want to do?

We can have a general discussion later. With other project like Retirment Hub we used MUI to handle component style and behavior. As for the layout we used Tailwind.

PR #205, opened 2024-01-29
AB#2659 -- RAOIDC integration (make RSA key config required)

gregory-j-baker:
Make RSA keys required at runtime so the application fails fast if they are missing. We don't want to provide default keys for security reasons.

## Related ADO workitems
- [AB#2659](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/2659)

### DIFF
--- a/frontend/.env.example
+++ b/frontend/.env.example
@@ -1,21 +1,24 @@
 # Example .env configuration file.
 # Copy this file to `.env` and edit as required.
 
+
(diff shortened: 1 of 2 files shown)


nicholas-ly (approved):

gregory-j-baker:
Note: waiting for #210 to merge so I can get the tests in this branch working.

sebastien-comeau (approved):

PR #207, opened 2024-01-29
Add InputRadios component

sebastien-comeau:
As per title. Unit tests will be created next.

### DIFF
--- a/frontend/app/routes/_gcweb-app.personal-information.preferred-language.edit.tsx
+++ b/frontend/app/routes/_gcweb-app.personal-information.preferred-language.edit.tsx
@@ -4,6 +4,7 @@ import { Form, Link, useLoaderData } from '@remix-run/react';
 import { useTranslation } from 'react-i18next';
 import { z } from 'zod';
 
+import { InputRadios } from '~/components/input-radios';
 import { lookupService } from '~/services/lookup-service.server';
 import { sessionService } from '~/services/session-service.server';
 import { userService } from '~/services/user-service.server';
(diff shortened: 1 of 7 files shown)


frontend/app/routes/_gcweb-app.personal-information.preferred-language.edit.tsx
```diff
@@ -67,20 +68,19 @@ export default function PreferredLanguageEdit() {
...
+          id="preferred-language"
+          name="preferredLanguage"
+          legend={t('personal-information:preferred-language.edit.page-title')}
+          options={
+            preferredLanguageList?.map((language) => ({
+              defaultChecked: userInfo.preferredLanguage === language.id,
+              children: getNameByLanguage(i18n.language, language),
+              value: language.id,
+            })) ?? []
+          }
```
gregory-j-baker, line 81: Wouldn't it be better to check if there is data in `preferredLanguageList` and then conditionally render the component based on that check?

gregory-j-baker (dismissed):
Approved with a comment.

frontend/app/routes/_gcweb-app.personal-information.preferred-language.edit.tsx
sebastien-comeau: It's "not" recommended by react to return null or empty fragment. It's the responsability of the consument to conditionally render the InputRadios component.

https://react.dev/learn/conditional-rendering#conditionally-returning-nothing-with-null

> In practice, returning null from a component isn’t common because it might surprise a developer trying to render it. More often, you would conditionally include or exclude the component in the parent component’s JSX.

gregory-j-baker: But in this case, the code **is in the parent**. The check can't be moved up any further in the component tree. (maybe I'm missing the point of your comment)

As coded, if there's no preferred languages, we will get:

```
Preferred language
{nothing}
```

Wouldn't it make more sense to either
a) not render the entire `<InputRadios>` component, or
b) throw

I don't see any reason to render a label when there are no values to select.

sebastien-comeau: I get what you mean. Sorry I thought you were put that logic in `InputRadios` component. I can conditionally render the InputRadios component.

sebastien-comeau: Fixed!

gregory-j-baker (approved):

nicholas-ly (approved):

PR #209, opened 2024-01-29
Lazy-load redis environment variables

gregory-j-baker:
The redis environment variables were being loaded whenever the module was being imported, even if redis wasn't configured as the session store.


Fbasham (approved):
pulled and tested. Works fine.

PR #210, opened 2024-01-29
Fix mocks in personal-information tests

gregory-j-baker:
Some unit tests were not mocking the session service, making the tests fragile.


Fbasham (approved):
i totally pulled this and ran it locally...

PR #217, opened 2024-01-30
Change required asterisk css font-size

sebastien-comeau:
Change required asterisk css font-size to `1rem` otherwise it match the parrent element and can be outside of the margin of the content.

image


gregory-j-baker:
Got a screenshot of after the change?

sebastien-comeau:
> Got a screenshot of after the change?

image

PR #241, opened 2024-02-01
AB#2886 -- Creating fake address entities and mocking endpoint for getting an address

nicholas-ly:
### Description
- Using faker to create two new address entities in `mocks/db.ts` to represent a user's home and mailing address
- Exposing an example endpoint that retrieves a specific address given an address identifier in `mocks/node.ts`

[AB#2886](https://dev.azure.com/DTS-STN/Canada%20Dental%20Care%20Plan/_workitems/edit/2886)

### Checklist
- [x] I have tested the changes locally.
- [x] My code follows the project's coding style.
- [x] I have updated the documentation if necessary.
- [x] All new and existing tests passed.

### Test Instructions
This is an incremental PR and as such, the only way to confirm it works is by checking that the address identifiers display in the Address panel on `/personal-information`.

### Additional Notes
This PR will result in the address identifier being shown instead of the previous home and mailing address strings. This will be resolved in future PRs.

### DIFF
--- a/frontend/app/mocks/node.ts
+++ b/frontend/app/mocks/node.ts
@@ -27,7 +27,32 @@ function getUserEntity(id: string | readonly string[]) {
 }
 
 /**
- * Retrieves a preferred language entity based on the provided user ID.
+ * Retrieves an address entity based on the provided user ID and address ID.
+ *
+ * @param userId - The user ID to look up in the database.
+ * @param addressId - The address ID to look up in the database.
+ * @returns The address entity if found, otherwise throws a 404 error.
+ */
+function getAddressEntity(userId: string | readonly string[], addressId: string | readonly string[]) {
+  const userEntity = getUserEntity(userId);
+  const parsedAddressId = z.string().safeParse(addressId);
+
+  if (!parsedAddressId.success || ![userEntity.homeAddress, userEntity.mailingAddress].includes(parsedAddressId.data)) {
+    throw new HttpResponse('No address found', { status: 404, headers: { 'Content-Type': 'text/plain' } });
+  }
+  const parsedAddress = db.address.findFirst({
+    where: { id: { equals: parsedAddressId.data } },
+  });
+
+  if (!parsedAddress) {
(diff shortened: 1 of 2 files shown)


sebastien-comeau (commented):
Use [one-to-one](https://github.com/mswjs/data?tab=readme-ov-file#one-to-one) or [nested-structures](https://github.com/mswjs/data?tab=readme-ov-file#nested-structures) with `@mswjs/data`

nicholas-ly:
> Use [one-to-one](https://github.com/mswjs/data?tab=readme-ov-file#one-to-one) or [nested-structures](https://github.com/mswjs/data?tab=readme-ov-file#nested-structures) with `@mswjs/data`

Thanks. I will take advantage of this suggestion in a future PR.
Introducing one of these features into this PR will break the existing schema validation in `user-service.server.ts` as that service currently expects a string for home and mailing address rather than an (nested) object.

nicholas-ly:
> Sometimes you have to broke and fix. Especially if you are at the bottom of the stack.

I would fix in this PR but that would mean introducing/modifying the existing address schema within the `userInfoSchema`, which will introduce changes to the entire address workflow. Before you know it, all the tests will be broken too. I prefer the incremental PR approach to keep the changes small.

frontend/app/mocks/node.ts
```diff
@@ -86,6 +107,22 @@ const handlers = [
     return HttpResponse.text(null, { status: 204 });
   }),
 
+  /**
+   * Handler for GET request to retrieve address by id
+   */
+  http.get('https://api.example.com/users/:userId/addresses/:addressId', ({ params }) => {
```
sebastien-comeau, line 113: You must use the `userId` param to fetch the user first then use that user to fetch his related address.

sebastien-comeau (changes requested):

frontend/app/mocks/node.ts
nicholas-ly: Fixed in 2d834de935ac55c6e3fd5b53090e72574032ff3b

sebastien-comeau (approved):

PR #246, opened 2024-02-02
2871 create initial flow for view letter

shawn320:
### Description
Created the initial flow for viewing letters. Created a new route view-letter to render a list of letters. Added temporary dummy data to map out a list of letters, this will be replaced once the mock and the service are ready. This PR does not include the functionality to view and download PDFs.

- [2871](https://dev.azure.com/DTS-STN/Canada%20Dental%20Care%20Plan/_sprints/taskboard/LiteBrite/Canada%20Dental%20Care%20Plan/Sprint%202?workitem=2871)
- [2872](https://dev.azure.com/DTS-STN/Canada%20Dental%20Care%20Plan/_sprints/taskboard/LiteBrite/Canada%20Dental%20Care%20Plan/Sprint%202?workitem=2872)

Capture

### Checklist

- [x] I have tested the changes locally.
- [x] My code follows the project's coding style.
- [ ] I have updated the documentation if necessary.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] All new and existing tests passed.


frontend/app/components/letter-list.tsx
```diff
@@ -0,0 +1,58 @@
...
+  id: string;
+  subject: string;
+  dateSent: string;
+  referenceId: string;
+}
+
+export interface LetterListProps {
+  letters: Letter[];
+  i18n: { subject: string; dateSent: string; referenceId: string };
+}
```
gregory-j-baker, line 15:
```suggestion
export interface LetterListProps {
  letters: Letter[];
  subject: string;
  dateSent: string;
  referenceId: string;
}
```

Or, maybe have the component look up these strings itself (insead of passing them as props).

frontend/app/components/letter-list.tsx
gregory-j-baker, line 1: Does this react component need to exist outside of the page that uses it? Typically we pull components out into the `components/` folder when they are reusable. If a component will only ever be used in one page, I think it makes sense to just keep that component in the page's .tsx file.

gregory-j-baker: As an example, see `_gcweb-app.tsx`. That route contains a bunch of components, but because they'll only ever be used by that specific file, they're kept in that file.

frontend/app/components/letter-list.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const [isSortedLatest, setIsSortedLatest] = useState<boolean>(true);
+  const [sorted, setSorted] = useState<Letter[]>(letters);
+
+  useEffect(() => {
+    setSorted(sortLetters(letters, isSortedLatest));
+  }, [letters, isSortedLatest]);
+
+  const sortLetters = (letters: Letter[], latest: boolean) => {
+    return [...letters].sort((a, b) => {
+      return latest ? new Date(b.dateSent).getTime() - new Date(a.dateSent).getTime() : new Date(a.dateSent).getTime() - new Date(b.dateSent).getTime();
```
gregory-j-baker, line 27: Not sure, but I don't think you need to use `getTime()` here.

``` shell
❯ node
Welcome to Node.js v20.10.0.
Type ".help" for more information.
> const a = new Date()
> const b = new Date()
> b - a
4239 # ← 4.2s
```

gregory-j-baker (changes requested):

frontend/app/components/letter-list.tsx
shawn320: Will get an error if I remove `getTime()`.

`The left-hand side of an arithmetic operation must be of type 'any', 'number', 'bigint' or an enum type.ts(2362)`

However, this doens't break the app.
It's suggesting to use getTime() to converts date object into numeric timestamps, it's more explicit, makes it more accurate.

PR #255, opened 2024-02-05
refactor view-letter page

shawn320:
### Description
Remove dynamic route to view letter, instead of stream pdf content with <iframe>, download letter direcly. 

### Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have checked that all e2e tests pass by running `npm run test:e2e -- run`


frontend/app/routes/_gcweb-app.view-letters._index.tsx
```diff
@@ -58,6 +59,7 @@ interface LetterListProps {
 }
 
 function LetterList({ letters }: LetterListProps) {
+  const { filePath } = useLoaderData<LoaderData>();
```
nicholas-ly, line 62:
```suggestion
  const { filePath } = useLoaderData<<typeof loader>>();
```

This suggestion will remove the need to define the `LoaderData` type above.

nicholas-ly (dismissed):

frontend/app/routes/_gcweb-app.view-letters._index.tsx
sebastien-comeau, line 1: Change that route to `_gcweb-app.letters._index.tsx`. This route will list the current user's issued letters.
For the issued letters file download we can create the route `_gcweb-app.letters.$id.download.tsx` and this route will only have a loader to fetch and stream the file from the API using the letter's reference Id.

frontend/app/routes/_gcweb-app.view-letters._index.tsx
```diff
@@ -17,10 +17,12 @@ export const handle = {
 } as const satisfies RouteHandleData;
 
 // TODO Create loader function to fetch data from mock using service
+export async function loader({ request }: LoaderFunctionArgs) {
```
sebastien-comeau, line 20: Loader function is reponsible to fetch the data with sorting and paging. This is done on the server side. Client component will render based on the loader returned state (with useLoaderData).

sebastien-comeau (changes requested):

frontend/app/routes/_gcweb-app.view-letters._index.tsx
```diff
@@ -72,10 +70,25 @@ function LetterList({ letters }: LetterListProps) {
     });
   };
 
+  function formatDate(dateStr: string) {
```
sebastien-comeau, line 73: This is not required since the dates coming from your letters mock are already formatted correctly. BUT in a future interation, dates coming from the API would surelly be in UTC format `"1996-02-02T03:04:05.000Z"` and we would need a way format it correctly based on the client local and timezone, not the server's local and timezone.

frontend/app/routes/_gcweb-app.view-letters._index.tsx
```diff
@@ -72,10 +70,25 @@ function LetterList({ letters }: LetterListProps) {
...
+  }
+
   return (
     <div>
-      <button onClick={() => setSortOrder(sortOrder === 'ASC' ? 'DESC' : 'ASC')}>Sort by date</button>
-      <div className="border-l border-r border-t">
+      <label htmlFor="sortOrder" className="text-sm">
+        <strong>{t('view-letters:index.filter')}</strong>
+      </label>
+      <select id="sortOrder" value={sortOrder} onChange={(e) => setSortOrder(e.target.value as 'ASC' | 'DESC')} className="text-sm">
```
sebastien-comeau, line 87: On select change event use `useNavigate()` hook to add search param `sort` with the select value. The loader will react on the new request and return the list sorted accordingly. If no `sort` pass to the loader then sort by `DESC` by default.

sebastien-comeau (approved):

PR #258, opened 2024-02-06
AB#2880, AB#2889: Mock Letter and PDF endpoints

carolinegeorge-gov:
### Description

-  Created mock PowerPlatform endpoint for retrieving letters
- Created mock CCT endpoint for creating an empty PDF
- Added CCT url to environment variables
- Created `letters-service` and `pdf-service` 

[AB#2880](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/2880): Mock Letter endpoint 
[AB#2889](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/2889): Mock PDF endpoint

### Checklist

- [x] I have checked that all e2e tests pass by running `npm run test:e2e -- run`

### Additional Notes

Assumptions were made about the letters and PDF response schema based off the information provided in the tasks. Please validate any information about it or the flow as necessary.

### DIFF
--- a/frontend/app/mocks/node.ts
+++ b/frontend/app/mocks/node.ts
@@ -73,7 +73,7 @@ function getPreferredLanguageEntity(id: string | readonly string[]) {
 }
 
 /**
- * Retrieves letter entities based on the provided user ID.
+ * Retrieves list of letter entities based on the provided user ID.
  *
  * @param userId - The user ID to look up in the database.
  * @returns The letter entity if found, otherwise throws a 404 error.
@@ -94,6 +94,27 @@ function getLetterEntities(userId: string | readonly string[]) {
   }
 
   return letterEntity;
+}
+
+/**
+ * Retrieves a PDF entity based on the provided referenceId ID.
+ *
+ * @param id - The reference Id to look up in the database.
+ * @returns The PDF entity if found, otherwise throws a 404 error.
+ */
+function getPdfEntity(referenceId: string | readonly string[]) {
+  const parsedReferenceId = z.string().safeParse(referenceId);
+  const parsedPdfEntity = !parsedReferenceId.success
+    ? undefined
+    : db.pdf.findFirst({
+        where: { id: { equals: parsedReferenceId.data } },
+      });
+
+  if (!parsedPdfEntity) {
+    throw new HttpResponse('No PDF found', { status: 404, headers: { 'Content-Type': 'text/plain' } });
+  }
+
+  return parsedPdfEntity;
 }
 
 /**
@@ -248,7 +269,7 @@ const handlers = [
   /**
    * Handler for GET requests to retrieve letters details.
    */
-  http.get('https://api.example.com/letters/:userId', ({ params }) => {
+  http.get('https://api.example.com/letters?userId=', ({ params }) => {
     const letterEntities = getLetterEntities(params.userId);
 
     return HttpResponse.json(
@@ -260,6 +281,31 @@ const handlers = [
         };
       }),
     );
+  }),
+
+  /**
+   * Handler for GET requests to retrieve pdf
+   */
+  http.get('https://api.example.com/cct/letters/:referenceId', async ({ params }) => {
+    const encoder = new TextEncoder();
+    const pdfEntity = getPdfEntity(params.referenceId);
+    const stream = new ReadableStream({
+      start(controller) {
+        // Encode the string chunks using "TextEncoder".
+        controller.enqueue(encoder.encode('Gums'));
+        controller.enqueue(encoder.encode('n'));
+        controller.enqueue(encoder.encode('proses'));
+        controller.enqueue(encoder.encode(pdfEntity?.id));
+        controller.close();
+      },
+    });
+
+    // Send the mocked response immediately.
+    return new HttpResponse(stream, {
+      headers: {
+        'Content-Type': 'text/plain',
+      },
+    });
   }),
 ];
 
--- a/frontend/app/mocks/node.ts
+++ b/frontend/app/mocks/node.ts
@@ -81,7 +81,7 @@ function getPreferredLanguageEntity(id: string | readonly string[]) {
 function getLetterEntities(userId: string | readonly string[]) {
   const parsedUserId = z.string().uuid().safeParse(userId);
   if (!parsedUserId) {
-    throw new HttpResponse('Invalid userId: ' + parsedUserId, { status: 404, headers: { 'Content-Type': 'text/plain' } });
+    throw new HttpResponse('Invalid userId: ' + parsedUserId, { status: 400, headers: { 'Content-Type': 'text/plain' } });
   }
   const letterEntities = !parsedUserId.success
     ? undefined
@@ -290,10 +290,6 @@ const handlers = [
   /**
    * Handler for GET requests to retrieve pdf
    */
-
-  /**
-   *
-   */
   http.get('https://api.example.com/cct/letters/:referenceId', async ({ params }) => {
     const encoder = new TextEncoder();
     const pdfEntity = getPdfEntity(params.referenceId);
@@ -311,7 +307,8 @@ const handlers = [
     // Send the mocked response immediately.
     return new HttpResponse(stream, {
       headers: {
-        'Content-Type': 'text/plain',
+        'Content-Type': 'application/octet-stream',
+        'Content-Disposition': `attachment, filename=${pdfEntity.referenceId}.pdf`,
       },
     });
   }),
--- a/frontend/app/mocks/node.ts
+++ b/frontend/app/mocks/node.ts
@@ -73,6 +73,30 @@ function getPreferredLanguageEntity(id: string | readonly string[]) {
 }
 
 /**
+ * Retrieves letter entities based on the provided user ID.
+ *
+ * @param userId - The user ID to look up in the database.
+ * @returns The letter entity if found, otherwise throws a 404 error.
+ */
+function getLetterEntities(userId: string | readonly string[]) {
+  const parsedUserId = z.string().uuid().safeParse(userId);
+  const letterEntity = !parsedUserId.success
+    ? undefined
+    : db.letter.findMany({
+        where: {
+          userId: {
+            equals: parsedUserId.data,
+          },
+        },
+      });
+  if (!letterEntity) {
+    throw new HttpResponse('Letter Not found', { status: 404, headers: { 'Content-Type': 'text/plain' } });
+  }
+
+  return letterEntity;
+}
+
+/**
  * Converts a user entity to a patch document with specific fields.
  *
  * @param userEntity - The user entity to convert.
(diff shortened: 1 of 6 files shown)


frontend/app/mocks/node.ts
```diff
@@ -72,6 +72,51 @@ function getPreferredLanguageEntity(id: string | readonly string[]) {
...
+    : db.pdf.findFirst({
+        where: { id: { equals: parsedReferenceId.data } },
+      });
+
+  if (!parsedPdfEntity) {
+    throw new HttpResponse('No PDF found', { status: 404, headers: { 'Content-Type': 'text/plain' } });
+  }
+
+  return parsedPdfEntity;
+}
```
gregory-j-baker, line 118:
```suggestion
function getPdfEntity(referenceId: string | readonly string[]) {
  const parsedReferenceId = z.string().min(1).safeParse(referenceId);

  if (!parsedReferenceId.success) {
    throw new HttpResponse('Invalid reference id', { status: 400 });
  }

  const pdfEntity = db.pdf.findFirst({
    where: {
      id: { equals: parsedReferenceId.data },
    },
  });

  if (!pdfEntity) {
    throw new HttpResponse(null, { status: 404 });
  }

  return pdfEntity;
}
```

**Reason:** be explicit about what the function is doing. If input validation fails, then let the caller know by responding with a 400.

frontend/app/mocks/node.ts
```diff
@@ -72,6 +72,51 @@ function getPreferredLanguageEntity(id: string | readonly string[]) {
...
+  return letterEntity;
+}
+
+/**
+ * Retrieves a PDF entity based on the provided referenceId ID.
+ *
+ * @param id - The reference Id to look up in the database.
+ * @returns The PDF entity if found, otherwise throws a 404 error.
+ */
+function getPdfEntity(referenceId: string | readonly string[]) {
```
gregory-j-baker, line 105: Why can `userId` be an array of strings here? I feel like it makes sense to limit this to just a single user id.

frontend/app/mocks/node.ts
```diff
@@ -168,6 +213,48 @@ const handlers = [
...
+        controller.enqueue(encoder.encode('proses'));
+        controller.enqueue(encoder.encode(pdfEntity?.id));
+        controller.close();
+      },
+    });
+
+    // Send the mocked response immediately.
+    return new HttpResponse(stream, {
+      headers: {
+        'Content-Type': 'text/plain',
```
gregory-j-baker, line 254: I could be wrong about this, but shouldn't this be `application/octet-stream`?

frontend/app/services/letters-service.server.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    throw new Error(`Failed to fetch data. Status: ${response.status}, Status Text: ${response.statusText}`);
+  }
+
+  return {
+    getLetters,
+  };
+}
+
+export const lettersService = createLettersService();
```
gregory-j-baker, line 46: Take a look at `raoidc-service.server.ts` file and follow that approach to create a singleton service (ie: use `moize`). Eventually all of the services will be refactored to use that approach.

frontend/app/services/letters-service.server.ts
```diff
@@ -0,0 +1,46 @@
...
+  const logger = getLogger('letters-service.server');
+  const { INTEROP_API_BASE_URI } = getEnv();
+
+  /**
+   *
+   * @param { userId }
+   * @returns returns the letters based off  @param userId
+   */
+  async function getLetters(requestBody: { userId: string }) {
+    const url = `${INTEROP_API_BASE_URI}/letters?userId=${requestBody.userId}`;
```
gregory-j-baker, line 24:
```suggestion
    const url = new URL(`${INTEROP_API_BASE_URI}/letters`);
    url.searchParams.set('userId', requestBody.userId);
```

frontend/app/services/letters-service.server.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    logger.error('%j', {
+      message: 'Failed to fetch data',
+      status: response.status,
+      statusText: response.statusText,
+      url: url,
+      responseBody: await response.text(),
+    });
+
+    throw new Error(`Failed to fetch data. Status: ${response.status}, Status Text: ${response.statusText}`);
```
gregory-j-baker, line 38: I would handle the error state earlier in the function. If there's a problem with the response, return early.

This then gives you the freedom afterwards to post-process the response data however you see fit. Granted, in this case, there's no post-processing happening.. but who knows what future iterations of this code might look like.

frontend/app/services/letters-service.server.ts
```diff
@@ -0,0 +1,46 @@
...
+   *
+   * @param { userId }
+   * @returns returns the letters based off  @param userId
+   */
+  async function getLetters(requestBody: { userId: string }) {
+    const url = `${INTEROP_API_BASE_URI}/letters?userId=${requestBody.userId}`;
+    const response = await fetch(url);
+
+    if (response.ok) return letterSchema.parse(await response.json());
+    if (response.status === 404) return null;
```
gregory-j-baker, line 28: Interop should return an empty array if no letters are found. If we get a 404 back from interop for this call, it's almost certainly a misconfiguration on our end, so we should throw.

In other words, don't add any special handling for the 404.. let your `!response.ok` check handle it.

frontend/app/services/pdf-service.ts
```diff
@@ -0,0 +1,46 @@
...
+  const logger = getLogger('pdf-service.server');
+  const { CCT_API_BASE_URI } = getEnv();
+
+  /**
+   *
+   * @param { referenceId }
+   * @returns returns the pdf based off  @param referenceId
+   */
+  async function getPdf(referenceId: string) {
+    const url = `${CCT_API_BASE_URI}/cct/letters/${referenceId}`;
```
gregory-j-baker, line 24:
```suggestion
    const url = new URL(referenceId, `${CCT_API_BASE_URI}/cct/letters/`);
```

frontend/app/services/pdf-service.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    logger.error('%j', {
+      message: 'Failed to fetch data',
+      status: response.status,
+      statusText: response.statusText,
+      url: url,
+      handlersresponseBody: await response.text(),
+    });
+
+    throw new Error(`Failed to fetch data. Status: ${response.status}, Status Text: ${response.statusText}`);
```
gregory-j-baker, line 38: As with the other service, I would recommend handling the error state early, so you leave room in the function to do any post-processing of data (should you need to).

gregory-j-baker: So in this case, you would do something like:

``` typescript
if (!response.ok) {
  if (response.status === 404) { return null; }
  // throw...
}

// rest of function
```

gregory-j-baker (changes requested):
For your future PRs, I think it would be better if you split them up so that they're a little smaller.

For example, for this PR you could have submitted one for the letter service, and then another for the PDF service.

If you wanted to make it even easier for reviewers, you could have further split the PRs into mocks, then routes.

Doing this helps reviewers, but also limits the amount of copy/paste issues that arise. Some of the comments I made here were because both the letter service and PDF service were very similar implementations (ie: they seems to be copy/pastes of each other).

frontend/app/mocks/node.ts
```diff
@@ -72,6 +72,51 @@ function getPreferredLanguageEntity(id: string | readonly string[]) {
   return parsedPreferredLanguage;
 }
 
+/**
+ * Retrieves list of letter entities based on the provided user ID.
+ *
+ * @param userId - The user ID to look up in the database.
+ * @returns The letter entity if found, otherwise throws a 404 error.
+ */
+function getLetterEntities(userId: string | readonly string[]) {
```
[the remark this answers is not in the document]
sebastien-comeau: @gregory-j-baker because handler `param` values is type of `string | readonly string[]`. We could move the `param` value check in the handler and have explicity `getEntity` argument's type. Force the consument to pass the proper arguement's type.

frontend/app/mocks/node.ts
```diff
@@ -72,6 +72,51 @@ function getPreferredLanguageEntity(id: string | readonly string[]) {
...
+            equals: parsedUserId.data,
+          },
+        },
+      });
+  if (!letterEntity) {
+    throw new HttpResponse('Letter Not found', { status: 404, headers: { 'Content-Type': 'text/plain' } });
+  }
+
+  return letterEntity;
+}
```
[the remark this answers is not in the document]
carolinegeorge-gov: Does that include if its empty? Or will we show the page here regardless?

frontend/app/services/pdf-service.ts
sebastien-comeau: This could be applied as a service refactoring task.

frontend/app/mocks/node.ts
gregory-j-baker: Sorry, I'm not sure I understand your point here.

This is mocking an API which is supposed to be returning a bytestream. I suppose we *could* designate the type as `application/pdf`, but definitely not `text/plain`.

If it's helpful, Box.com's file download API returns its bytestream as `application/octet-stream`: <https://developer.box.com/reference/get-files-id-content/#response>

frontend/app/services/pdf-service.ts
gregory-j-baker: Are you suggesting in a future PR?

frontend/app/mocks/node.ts
sebastien-comeau: If the mock is returning a PDF then yes... but for not it's returning plain text content. Ok to switch it to `application/octet-stream` based on Box Platform.

frontend/app/services/pdf-service.ts
sebastien-comeau: > Are you suggesting in a future PR?

Yes

frontend/app/mocks/node.ts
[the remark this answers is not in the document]
carolinegeorge-gov: Fixed: 47fffce2dc8f67ae5494d7842d37f6acb35a77cc

frontend/app/mocks/node.ts
carolinegeorge-gov: Fixed: 47fffce2dc8f67ae5494d7842d37f6acb35a77cc

frontend/app/services/letters-service.server.ts
carolinegeorge-gov: Since the other services use the former format, I want to keep it consistent. Otherwise, I feel it should be changed all at the same time.

frontend/app/services/letters-service.server.ts
carolinegeorge-gov: FIxed: 47fffce2dc8f67ae5494d7842d37f6acb35a77cc

frontend/app/mocks/node.ts
sebastien-comeau: Use this header for now. Replace `{letter.referenceId}` is a placeholder and should be provider by the letter entity.
```
Content-Type: application/octet-stream
Content-Disposition: attachment; filename="{letter.referenceId}.pdf"
```

frontend/app/services/letters-service.server.ts
sebastien-comeau: Other services doens't have `searchParams` change this one and we can refactor the other places in another PR.

frontend/app/mocks/node.ts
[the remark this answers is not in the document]
gregory-j-baker: I hate nitpicking this since it's only a mock and we have no idea how the actual letter API will work.. but this should probably be a `400` instead of a `404`.

frontend/app/services/letters-service.server.ts
gregory-j-baker: That's fair. I'll resolve the comment.

frontend/app/mocks/node.ts
```diff
@@ -216,15 +262,59 @@ const handlers = [
...
+    );
+  }),
+
+  /**
+   * Handler for GET requests to retrieve pdf
+   */
+
+  /**
+   *
+   */
```
gregory-j-baker, line 296: This can be removed.

gregory-j-baker (changes requested):

frontend/app/mocks/node.ts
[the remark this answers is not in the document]
carolinegeorge-gov: Good point. Fixed: c20e06602530ac9a3fb8cf2b080a3b3fc36d0451

frontend/app/mocks/node.ts
carolinegeorge-gov: Fixed: c20e06602530ac9a3fb8cf2b080a3b3fc36d0451

frontend/app/services/letters-service.server.ts
carolinegeorge-gov: Got it. Fixed: c20e06602530ac9a3fb8cf2b080a3b3fc36d0451.

frontend/app/mocks/node.ts
carolinegeorge-gov: FIxed: c20e06602530ac9a3fb8cf2b080a3b3fc36d0451.

PR #260, opened 2024-02-06
Fix copy/paste error in pull request template

gregory-j-baker:
### Description

As per title.


sebastien-comeau (approved):
You had one job and you fixed it.

PR #271, opened 2024-02-07
AB#2770 -- Adding InputSelect and InputOption components

nicholas-ly:
### Description
This PR introduces re-usable and accessible components for `<select>` and `<option>` and is based off similar form components that currently exist in the application.

The corresponding components, `InputSelect` and `InputOption`, are will eventually be used to render a drop-down of countries and provinces/states within the home and mailing address forms (among other forms).

[AB#2770](https://dev.azure.com/DTS-STN/Canada%20Dental%20Care%20Plan/_workitems/edit/2770)

image

### Checklist
- [ ] I have added tests that prove my fix is effective or that my feature works

### Test Instructions
Run the application locally and navigate to `/personal-information/home-address/edit`.
The "Country" field uses a `InputSelect`.

### Additional Notes
Unit tests will be added in a subsequent PR.

### DIFF
--- a/frontend/app/components/input-option.tsx
+++ b/frontend/app/components/input-option.tsx
@@ -1,0 +1,18 @@
+import { type ComponentProps } from 'react';
+
+export interface InputOptionProps extends ComponentProps<'option'> {
+  id: string;
+  label: string;
+  value: string;
+}
+
+export function InputOption(props: InputOptionProps) {
+  const { label, value, id, ...restProps } = props;
+  const inputOptionTestId = `input-option-${id}-test`;
+
+  return (
+    <option value={value} id={id} data-testid={inputOptionTestId} {...restProps}>
+      {label}
+    </option>
+  );
+}
--- a/frontend/app/components/input-select.tsx
+++ b/frontend/app/components/input-select.tsx
@@ -1,0 +1,80 @@
+import { type ComponentProps, type ReactNode, forwardRef } from 'react';
+
+import clsx from 'clsx';
+
+import { InputError } from './input-error';
+import { InputHelp } from './input-help';
+import { InputLabel } from './input-label';
+import { InputOption } from './input-option';
+
+export interface InputSelectProps extends Omit<ComponentProps<'select'>, 'aria-describedby' | 'aria-errormessage' | 'aria-invalid' | 'aria-labelledby' | 'aria-required'> {
+  errorMessage?: string;
+  helpMessage?: ReactNode;
+  helpMessageSecondary?: ReactNode;
+  id: string;
+  label: string;
+  name: string;
+  options: ComponentProps<typeof InputOption>[];
+}
+
+const InputSelect = forwardRef<HTMLSelectElement, InputSelectProps>((props, ref) => {
+  const { errorMessage, helpMessage, helpMessageSecondary, id, label, options, className, required, ...restInputProps } = props;
+
+  const inputErrorId = `input-${id}-error`;
+  const inputHelpMessageId = `input-${id}-help`;
+  const inputHelpMessageSecondaryId = `input-${id}-help-secondary`;
+  const inputLabelId = `input-${id}-label`;
+  const inputTestId = `input-${id}-test`;
+  const inputWrapperId = `input-${id}`;
+
+  function getAriaDescribedby() {
+    const ariaDescribedby = [];
+    if (helpMessage) ariaDescribedby.push(inputHelpMessageId);
+    if (helpMessageSecondary) ariaDescribedby.push(inputHelpMessageSecondaryId);
+    return ariaDescribedby.length > 0 ? ariaDescribedby.join(' ') : undefined;
+  }
+
+  return (
+    <div id={inputWrapperId} data-testid={inputWrapperId} className="form-group">
+      <InputLabel id={inputLabelId} htmlFor={id} required={required}>
+        {label}
+      </InputLabel>
+      {errorMessage && (
+        <InputError id={inputErrorId} className="mb-1.5">
+          {errorMessage}
+        </InputError>
+      )}
+      {helpMessage && (
+        <InputHelp id={inputHelpMessageId} className="mb-1.5" data-testid="input-select-help">
+          {helpMessage}
+        </InputHelp>
+      )}
+      <select
+        ref={ref}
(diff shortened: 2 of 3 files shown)


frontend/app/components/input-option.tsx
```diff
@@ -0,0 +1,16 @@
...
+export interface InputOptionProps extends ComponentProps<'option'> {
+  label: string;
+  value: string;
+}
+
+export function InputOption(props: InputOptionProps) {
+  const { label, value, ...restProps } = props;
+
+  return (
+    <option value={value} data-testid="input-radio" {...restProps}>
```
gregory-j-baker, line 12: - `data-testid` needs to be unique here
- this isn't a radio input 🤣

Maybe require a mandatory ID be passed in to the select component, and then pass that ID down to the option component and set the test id to `option-${id}` or something like that.

frontend/app/components/input-select.tsx
```diff
@@ -0,0 +1,79 @@
+import { type ComponentProps, type ReactNode, forwardRef } from 'react';
+
+import clsx from 'clsx';
+
+import { InputError } from './input-error';
+import { InputHelp } from './input-help';
+import { InputLabel } from './input-label';
+import { InputOption } from './input-option';
```
gregory-j-baker, line 8: Should these be aliased imports (`~/components/`)?

Maybe we should have a discussion about this, because I feel like in some situations, using `./` makes more sense.

frontend/app/components/input-select.tsx
```diff
@@ -0,0 +1,79 @@
...
+        aria-labelledby={inputLabelId}
+        aria-required={required}
+        className={clsx('form-control', className)}
+        data-testid="input-select"
+        id={id}
+        required={required}
+        {...restInputProps}
+      >
+        {options.map(({ label, value, ...restProps }) => (
+          <InputOption label={label} value={value} key={value} {...restProps} />
```
gregory-j-baker, line 65: Pass the id into `<InputOption>` so it can be mapped to the testid for the option element.

gregory-j-baker (changes requested):

frontend/app/components/input-select.tsx
sebastien-comeau:
```suggestion
          <InputOption key={value} {...restProps} />
```

frontend/app/components/input-option.tsx
sebastien-comeau: I disagre... the `InputOption` component consumer can override the `data-testid` if he pass it as a prop.

frontend/app/components/input-select.tsx
nicholas-ly: Since the components in question are closely related (and in the same directory) as the `InputSelect` component, I suggest leaving these lines as is.

frontend/app/components/input-select.tsx
nicholas-ly: @sebastien-comeau Since `label` and `value` are currently mandatory (as defined by `InputOptionProps`), we have to explicitly set them when using `<InputOption />`. Thoughts?

frontend/app/components/input-option.tsx
nicholas-ly: > I disagre... the `InputOption` component consumer can override the `data-testid` if he pass it as a prop. But th testid is wrong though. It should be `data-testid="input-option"`.

That was my assumption when creating this component. I will fix the `data-testid` value though.

gregory-j-baker: The wrapping component currently has no way to pass a test id down to the option component. And I still assert that these IDs need to be unique, even when the wrapping component omits the testid. This is true for both `<InputSelect>` and `<InputOption>`.

**Re: using the value as the id:** sure, that's okay.. but what if the value has spaces or other weird characters? Maybe you should hash the value? ex:

``` typescript
createHash('shake256', { outputLength: 4 }).update('value').digest('hex');
→ '457bdb6e'
```

frontend/app/components/input-select.tsx
gregory-j-baker: Yeah.. I wonder if we should create a `~/components/form` or similar folder to hold all of the form inputs?

Also, just noticed that `<InputOption>` is in its own file. Would you ever use this outside of the `<InputSelect>` component? If not, why not just move it inside?

nicholas-ly: > Yeah.. I wonder if we should create a `~/components/form` or similar folder to hold all of the form inputs?

I'm okay with that (in a future PR)

> Also, just noticed that `<InputOption>` is in its own file. Would you ever use this outside of the `<InputSelect>` component? If not, why not just move it inside?

My thinking was that if we ever need to use an `InputOption` inside some other element, like components to render a `<optgroup>` or `<datalist>`, having a separate `InputOption` file gives us this flexibility.

frontend/app/components/input-select.tsx
nicholas-ly: > @sebastien-comeau Since `label` and `value` are currently mandatory (as defined by `InputOptionProps`), we have to explicitly set them when using `<InputOption />`. Thoughts?

Correction: I had de-structured  `label` and `value` from `InputOptionProps`, meaning `label` was missing from `...restProps`.

`<InputOption>` now accepts an `id`, which is used to construct a value for `data-testId`.

frontend/app/components/input-option.tsx
nicholas-ly: `<InputOption>` now accepts an id, which is used to construct a value for `data-testId`.

gregory-j-baker: Passing the parent ID down to the option isn't quite enough. What if you have 10 option components? (they all have the ID of the parent select component)

And yes, I acknowledge that passing the ID down was in fact my original suggestion.

gregory-j-baker: Disregard my last comment, I see how you're doing it. Will resolve this thread.

frontend/app/components/input-select.tsx
gregory-j-baker: I didn't think about those use cases. I'll resolve this thread.

gregory-j-baker (approved):

PR #293, opened 2024-02-08
Tighten up envar validation (mocks)

gregory-j-baker:
### Description

Tighten envar validation so only valid mock names are allowed for `ENABLED_MOCKS`.

### Checklist

### Test Instructions

Add some funky values to your `ENABLED_MOCKS` envar. Try not setting it at all, even.

### DIFF
--- a/frontend/app/utils/env.server.ts
+++ b/frontend/app/utils/env.server.ts
@@ -12,13 +12,18 @@ function tryOrElseFalse(fn: () => unknown) {
   catch { return false; }
 }
 
+// @see app/mocks/node.ts
+const validMockNames = ['cct', 'lookup', 'power-platform', 'raoidc', 'wsaddress'];
+
 // refiners
+const areValidMockNames = (arr: Array<string>) => arr.every((str) => validMockNames.includes(str));
 const isValidPublicKey = (val: string) => tryOrElseFalse(() => publicKeyPemToCryptoKey(val));
 const isValidPrivateKey = (val: string) => tryOrElseFalse(() => privateKeyPemToCryptoKey(val));
 
 // transformers
+const csvToArray = (csv?: string) => csv?.split(',').map((str) => str.trim()) ?? [];
+const emptyToUndefined = (val?: string) => (val === '' ? undefined : val);
 const toBoolean = (val?: string) => val === 'true';
-const toUndefined = (val?: string) => (val === '' ? undefined : val);
 
 /**
  * Environment variables that will be available to server only.
@@ -34,7 +39,7 @@ const serverEnv = z.object({
   AUTH_JWT_PUBLIC_KEY: z.string().refine(isValidPublicKey),
   AUTH_RAOIDC_BASE_URL: z.string().trim().min(1),
   AUTH_RAOIDC_CLIENT_ID: z.string().trim().min(1),
-  AUTH_RAOIDC_PROXY_URL: z.string().trim().transform(toUndefined).optional(),
+  AUTH_RAOIDC_PROXY_URL: z.string().trim().transform(emptyToUndefined).optional(),
 
   // language cookie settings
   LANG_COOKIE_NAME: z.string().default('_gc_lang'),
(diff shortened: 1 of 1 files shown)


frontend/app/utils/env.server.ts
```diff
@@ -12,11 +12,16 @@ function tryOrElseFalse(fn: () => unknown) {
...
+// @see app/mocks/node.ts
+const mockNames = ['cct', 'lookup', 'power-platform', 'raoidc', 'wsaddress'];
+
 // refiners
+const isValidMockName = (val: Array<string>) => val.every((val) => mockNames.includes(val));
 const isValidPublicKey = (val: string) => tryOrElseFalse(() => publicKeyPemToCryptoKey(val));
 const isValidPrivateKey = (val: string) => tryOrElseFalse(() => privateKeyPemToCryptoKey(val));
 
 // transformers
+const toArray = (val?: string) => val?.split(',').map((val) => val.trim()) ?? [];
```
sebastien-comeau, line 24:
```suggestion
const csvToArray = (val?: string) => val?.split(',').map((val) => val.trim()) ?? [];
```

frontend/app/utils/env.server.ts
```diff
@@ -65,14 +70,8 @@ const serverEnv = z.object({
...
   // mocks settings
-  ENABLED_MOCKS: z
-    .string() // TODO :: GjB :: can this instead be a z.enum()
-    .transform((val) => val.split(',').map((val) => val.trim()))
-    .default(''),
-  MOCK_AUTH_ALLOWED_REDIRECTS: z
-    .string()
-    .transform((val) => val.split(',').map((val) => val.trim()))
-    .default('http://localhost:3000/auth/callback/raoidc'),
+  ENABLED_MOCKS: z.string().transform(toUndefined).transform(toArray).refine(isValidMockName).default(''),
```
sebastien-comeau, line 73: `ENABLED_MOCKS` type is `string[]` so why `.default('')`?

sebastien-comeau: Same for `MOCK_AUTH_ALLOWED_REDIRECTS`.

gregory-j-baker: These are envars, which are always simple strings (by definition).

In the case of these two vars, they're comma separated values.

``` properties
ENABLED_MOCKS=foo,bar,baz
MOCK_AUTH_ALLOWED_REDIRECTS=http://localhost:3000/auth/callback,https://example.com/auth
```

frontend/app/utils/env.server.ts
gregory-j-baker: I took your advice and renamed some of the variables to be more descriptive. Please re-review at your leisure.

frontend/app/utils/env.server.ts
```diff
@@ -65,14 +70,8 @@ const serverEnv = z.object({
...
   // mocks settings
-  ENABLED_MOCKS: z
-    .string() // TODO :: GjB :: can this instead be a z.enum()
-    .transform((val) => val.split(',').map((val) => val.trim()))
-    .default(''),
-  MOCK_AUTH_ALLOWED_REDIRECTS: z
-    .string()
-    .transform((val) => val.split(',').map((val) => val.trim()))
-    .default('http://localhost:3000/auth/callback/raoidc'),
+  ENABLED_MOCKS: z.string().transform(emptyToUndefined).transform(csvToArray).refine(areValidMockNames).default(''),
```
sebastien-comeau, line 73: Why not `.default([])`?

sebastien-comeau (approved):

frontend/app/utils/env.server.ts
sebastien-comeau: I read the Zod `.default` doccumentation.
https://zod.dev/?id=default

> Conceptually, this is how Zod processes default values:
> 
> 1. If the input is undefined, the default value is returned
> 2. Otherwise, the data is parsed using the base schema

gregory-j-baker: 🤷

PR #383, opened 2024-02-15
change content-type on pdf letter download

Fbasham:
This is a followup to #377 since I made and oopsie and forgot to change the content-type as well.  I've tested that the PDF now opens in all common browsers (Edge, FireFox, Chrome).


sebastien-comeau (approved):
Good catch. Normally we would use a utility function to get the proper MIME types based on the extension. This is something we can do later if ever the return file is something different than PDF.

PR #389, opened 2024-02-15
Added SCCH URLs env for dev and staging.

nibivi77:
### Description
Not sure if this is all the configuration needed.


gregory-j-baker (dismissed):
Our `dev` environment won't be integrated with SCCH/MSCAD.. can you revert that change and instead update the new `int` environment with the correct URL?

nicholas-ly:
> Our `dev` environment won't be integrated with SCCH/MSCAD.. can you revert that change and instead update the new `int` environment with the correct URL?

SCCH's dev environment (https://secure-client-hub-dev.bdm.dshp-phdn.net/) doesn't appear to require authentication, so why not use it instead of the current placeholder?

gregory-j-baker:
> > Our `dev` environment won't be integrated with SCCH/MSCAD.. can you revert that change and instead update the new `int` environment with the correct URL?
> 
> SCCH's dev environment (https://secure-client-hub-dev.bdm.dshp-phdn.net/) doesn't appear to require authentication, so why not use it instead of the current placeholder?

I didn't know this.. in that case, might as well configure our dev environment for S(M)C(S)C(C)H(AD)'s dev environment

nicholas-ly (approved):

PR #242, opened 2024-02-01
Add AnchorLink component

sebastien-comeau:
### Description
Add AnchorLink component used to create anchor links that scroll  and focus on the specified target element when clicked. Remix clear the client state if browser location change.

### Related Azure Boards Work Items

- [AB#2864](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/2864)

### Checklist

- [x] I have tested the changes locally.
- [x] My code follows the project's coding style.
- [x] I have updated the documentation if necessary.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] All new and existing tests passed.

### Test Instructions

Load the applicatin and navigate with the layout skip links and try the error summary links as well.

PR #401, opened 2024-02-16
remove preventDefault from anchor-link

Fbasham:
When using keyboard navigation and tabbing through the page, when the skip-link for navigating to `wb-cont` is pressed, nothing happens and the skip link remains in place.

On Canada.ca, if you follow the same procedure, the skip link correctly navigates to `wb-cont` and the skip-link disappears from view as expected.  

I'm not sure what the reason for having the `event.preventDefault()` is for in this case, but I believe it can be removed (or optionally toggled by adding a flag in the component props if needed).

### DIFF
--- a/frontend/app/components/anchor-link.tsx
+++ b/frontend/app/components/anchor-link.tsx
@@ -23,12 +23,10 @@ export function AnchorLink(props: AnchorLinkProps) {
 
   /**
    * handleOnSkipLinkClick is the click event handler for the anchor link.
-   * It prevents the default anchor link behavior, scrolls to and focuses
-   * on the target element specified by 'anchorElementId', and invokes
-   * the optional 'onClick' callback.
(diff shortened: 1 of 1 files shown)


sebastien-comeau:
If you don't prevent it the url changed as intended but it breaks Remix route states.

sebastien-comeau:
> If you don't prevent it the url changed as intended but it breaks Remix route states.

https://github.com/DTS-STN/canadian-dental-care-plan/pull/242

gregory-j-baker:
The `preventDefault()` call was added because without it, react router will perform a navigation when you invoke a frag link. This causes any client side state (ie: form validation data) to be lost. In my opinion, this react router behavior is a bug.

If keyboard navigation is broken for these links, I think the solution has to be to add the correct event handler for the keyboard event.

Fbasham:
closing...not worth the hassle. Thanks for the great DX remix

PR #443, opened 2024-02-20
Add buttons component unit tests

sebastien-comeau:
### Description

As per title.

### Checklist


frontend/__tests__/components/buttons.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+import { Button, ButtonLink } from '~/components/buttons';
+
+describe('Button Component', () => {
+  it('renders button with default props', () => {
+    const { getByText } = render(<Button>Click me</Button>);
+    const button = getByText('Click me');
+    expect(button).toBeInTheDocument();
+    expect(button.tagName).toBe('BUTTON');
+    expect(button).toHaveClass('inline-flex', 'items-center', 'justify-center', 'align-middle', 'font-medium', 'focus:outline-none', 'focus:ring-4');
```
gregory-j-baker, line 15: This assertion seems kind of fragile. Do we really care so much about the button styling that we need to check the css classes?

frontend/__tests__/components/buttons.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { createRemixStub } from '@remix-run/testing';
+
+import { describe, expect, it, vi } from 'vitest';
+
+import { Button, ButtonLink } from '~/components/buttons';
+
+describe('Button Component', () => {
+  it('renders button with default props', () => {
+    const { getByText } = render(<Button>Click me</Button>);
+    const button = getByText('Click me');
```
gregory-j-baker, line 12: Why aren't you using `data-testid` for this?

frontend/__tests__/components/buttons.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+    expect(button).toBeInTheDocument();
+    expect(button.tagName).toBe('BUTTON');
```
gregory-j-baker, line 14: You should use `.toEqual()` in every case instead of `.toBe()`

gregory-j-baker (changes requested):

frontend/__tests__/components/buttons.test.tsx
sebastien-comeau: https://vitest.dev/api/expect#tobe
> `toBe` can be used to assert if primitives are equal or that objects share the same reference. It is equivalent of calling `expect(Object.is(3, 3)).toBe(true)`. If the objects are not the same, but you want to check if their structures are identical, you can use [toEqual](https://vitest.dev/api/expect#toequal).

frontend/__tests__/components/buttons.test.tsx
sebastien-comeau: Since I have control over the button content I don't see the need of using `data-testid` and it assert that `children` is rendered correctly.

frontend/__tests__/components/buttons.test.tsx
sebastien-comeau: Those classes are important for the button base css classes.

frontend/__tests__/components/buttons.test.tsx
gregory-j-baker: Just always use `.toEqual()` since it works with both primitives *and* objects.

When you selectively use `.toBe()`, future developers can potentially copy/paste it into their tests without understanding the subtle difference. Using `.toEqual()` is always the safer choice, for current and future generations.

frontend/__tests__/components/buttons.test.tsx
sebastien-comeau: The rendered button have more classes than that.

frontend/__tests__/components/buttons.test.tsx
sebastien-comeau: `.toBe()` should be used in the majority of cases... `.toEqual()` should be use if you want to compare object. Devs should know the difference and if they not then they need to read the documentation.

frontend/__tests__/components/buttons.test.tsx
gregory-j-baker: 🤷

frontend/__tests__/components/buttons.test.tsx
gregory-j-baker: Sorry, but I'm being stubborn about this one..

This is in reference to Jest, but still applies to Vitest.

<https://dev.to/thejaredwilcurt/why-you-should-never-use-tobe-in-jest-48ca>
image

sebastien-comeau: Changed to `.toEqual()`

PR #420, opened 2024-02-16
AB#2705 -- Updating WSAddress mocks based on examples from Interop's OpenAPI specs

nicholas-ly:
### Description
This is an incremental pull request. `WSAddressService` will be broken in the meantime but I have commented out the only place where it is used with a `TODO`.

Also sneaked in a fix dealing with the mapping of the "Province/State" input options.

### Related Azure Boards Work Items
[AB#2705](https://dev.azure.com/DTS-STN/Canada%20Dental%20Care%20Plan/_workitems/edit/2705)

### Checklist
- [x] I have tested the changes locally
- [x] I have updated the documentation if necessary
- [x] I have added/updated tests that prove my fix is effective or that my feature works
- [x] I have checked that my code follows the project's coding style by running `npm run format:check`
- [x] I have checked that my code contains no linting errors by running `npm run lint`
- [x] I have checked that my code contains no type errors by running `npm run typecheck`
- [x] I have checked that all unit tests pass by running `npm run test:unit -- run`
- [x] I have checked that all e2e tests pass by running `npm run test:e2e`

### Additional Notes
This is an incremental pull request. `WSAddressService` must be updated to consume the mock correctly.

PR #450, opened 2024-02-20
AB#2705 -- Refactoring wsaddress-service to correctly consume (mocked) API

nicholas-ly:
### Description

This PR is a continuation of https://github.com/DTS-STN/canadian-dental-care-plan/pull/420.

#### `wsaddress-service.server.ts` changes:
- Functions now call `GET` with query parameters (as per mock API changes) instead of `POST` with request body
- Functions map from API's (NIEM'd) fields to a more developer-friendly domain entity
- Superfluous DTOs and Zod schemas have been removed

#### `wsaddress-service.server.test.ts` changes:
- Removing use of `handler`s and MSW's `setupServer(..)` as these uses are more suitable for integration tests rather than unit tests
- Mocking `fetch(..)` calls with (NIEM'd) responses
- Adding unit tests for when API response is not a `200 - OK` to obtain 100% unit test coverage 

[AB#2705](https://dev.azure.com/DTS-STN/Canada%20Dental%20Care%20Plan/_workitems/edit/2705)

### Checklist

### Test Instructions
Run the unit associated unit test for the service changes

### Additional Notes
Sorry for the large(r) pull request. Please let me know if you would like to break it down into smaller PRs.

### DIFF
--- a/frontend/app/services/wsaddress-service.server.ts
+++ b/frontend/app/services/wsaddress-service.server.ts
@@ -6,166 +6,34 @@ import { getLogger } from '~/utils/logging.server';
 
 const log = getLogger('wsaddress-service.server');
 
-export interface CorrectWSAddressRequestDTO {
-  addressLine: string;
-  city: string;
-  province: string;
-  postalCode: string;
-  country: string;
-  formatResult: boolean;
-  language: string;
-}
-
-export interface CorrectWSAddressResponseDTO {
-  responseType?: string;
-  addressLine?: string;
-  city?: string;
-  province?: string;
-  postalCode?: string;
-  deliveryInformation?: string;
-  extraInformation?: string;
-  statusCode?: string;
-  country?: string;
-  message?: string;
-  warnings?: string;
-  functionalMessages?: { action: string; message: string }[];
-}
-
-export interface ParseWSAddressRequestDTO {
-  addressLine: string;
-  city: string;
-  province: string;
-  postalCode: string;
-  country: string;
-  geographicScope: string;
-  parseType: string;
-  language: string;
-}
-
-export interface ParseWSAddressResponseDTO {
-  responseType?: string;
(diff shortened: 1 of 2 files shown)


frontend/app/services/wsaddress-service.server.ts
```diff
@@ -6,191 +6,65 @@ import { getLogger } from '~/utils/logging.server';
...
+        status: correctionResults['wsaddr:StatusCode'],
+        address: correctionResults['nc:AddressFullText'],
+        city: correctionResults['nc:AddressCityName'],
+        province: correctionResults['nc:ProvinceCode'],
+        postalCode: correctionResults['nc:AddressPostalCode'],
+        country: correctionResults['nc:CountryCode'],
+      };
     }
 
     log.error('%j', {
```
gregory-j-baker, line 70: I mentioned this in other PRs, and will mention it again.. I'm not **necessarily** suggesting a change, just want to start a discussion.

I feel like a better order to these services is:

1. fetch
1. handle errors
1. post-process and return

The reason being that sometimes we want to format, tweak, or do some other post-processing with the returned data. We're not doing that now, but we might in the future, and then we're going to end up with a huge `if { /* post-processing here */ }` block right in the middle of the function.

After writing a bunch of these over the years, I find that this is better:

``` typescript
function fetchSomething() {
  const resp = fetch(..);

  if (!resp.ok) {
    // throw or something
  }

  const data = resp.json();

  //
  // validate, map, combine, tweak... or do whatever with the data here
  //

  return retVal;
}
```

gregory-j-baker (dismissed):

frontend/app/services/wsaddress-service.server.ts
nicholas-ly: Good point and agreed.

gregory-j-baker (approved):

PR #455, opened 2024-02-21
AB#2935: add session timeout variables into env

leiye29:
### Description

1. add SESSION_TIMEOUT into env
2. add SESSION_TIMEOUT_PROMPT into env

### Checklist

### DIFF
--- a/frontend/.env.example
+++ b/frontend/.env.example
@@ -81,7 +81,7 @@ SESSION_COOKIE_SAME_SITE=lax
 SESSION_COOKIE_SECRET=0123456789ABCDEF
 # session cookie max age in seconds
 # (optional; default: 600)
-SESSION_COOKIE_MAX_AGE_SECONDS=600
+SESSION_COOKIE_MAX_AGE_SECONDS=1200
 # session cookie httpOnly flag
 # (optional; default: true)
 SESSION_COOKIE_HTTP_ONLY=true
@@ -91,7 +91,12 @@ SESSION_COOKIE_SECURE=true
 # location of session files when using file storage
 # (optional; default: ./node_modules/.cache/sessions/)
 SESSION_FILE_DIR=./node_modules/.cache/sessions/
-
+# the time before a user is considered idle in seconds
+# (optional; default: 1140)
+SESSION_TIMEOUT_SECONDS=1140
+# how long before idle to emit the onPrompt event in seconds
+# (optional; default: 300)
+SESSION_TIMEOUT_PROMPT_SECONDS=300
 
 # redis server URL
 # (optional; default: redis://localhost)
--- a/frontend/app/utils/env.server.ts
+++ b/frontend/app/utils/env.server.ts
@@ -67,6 +67,14 @@ const serverEnv = z.object({
   SESSION_COOKIE_HTTP_ONLY: z.string().transform(toBoolean).default('true'),
   SESSION_COOKIE_SECURE: z.string().transform(toBoolean).default('true'),
   SESSION_FILE_DIR: z.string().trim().min(1).default('./node_modules/cache/sessions/'),
+  SESSION_TIMEOUT_SECONDS: z.coerce
+    .number()
+    .min(0)
+    .default(19 * 60),
+  SESSION_TIMEOUT_PROMPT_SECONDS: z.coerce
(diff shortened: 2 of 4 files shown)


frontend/.env.example
```diff
@@ -91,7 +91,10 @@ SESSION_COOKIE_SECURE=true
 # location of session files when using file storage
 # (optional; default: ./node_modules/.cache/sessions/)
 SESSION_FILE_DIR=./node_modules/.cache/sessions/
-
+# SESSION_TIMEOUT=19 * 60 * 1000 (20 mins)
+# SESSION_TIMEOUT_PROMPT=5 * 60 *1000 (5 mins)
+SESSION_TIMEOUT=1140000
+SESSION_TIMEOUT_PROMPT=300000
```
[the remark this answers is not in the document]
leiye29: updated

frontend/app/utils/env.server.ts
```diff
@@ -67,6 +67,8 @@ const serverEnv = z.object({
   SESSION_COOKIE_HTTP_ONLY: z.string().transform(toBoolean).default('true'),
   SESSION_COOKIE_SECURE: z.string().transform(toBoolean).default('true'),
   SESSION_FILE_DIR: z.string().trim().min(1).default('./node_modules/cache/sessions/'),
+  SESSION_TIMEOUT: z.coerce.number().default(1140000),
+  SESSION_TIMEOUT_PROMPT: z.coerce.number().default(300000),
```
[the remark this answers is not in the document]
leiye29: updated

frontend/.env.example
```diff
@@ -91,7 +91,12 @@ SESSION_COOKIE_SECURE=true
 # location of session files when using file storage
 # (optional; default: ./node_modules/.cache/sessions/)
 SESSION_FILE_DIR=./node_modules/.cache/sessions/
-
+# the time before a user is considered idles in seconds
+# session-timeout.tsx
+SESSION_TIMEOUT_SECONDS=1140
+# how long before idle to emit the onPrompt event in seconds
+# session-timeout.tsx
+SESSION_TIMEOUT_PROMPT_SECONDS=300
```
nicholas-ly, line 99: **Nitpick:** To be consistent with the rest of the file

```suggestion
# the time before a user is considered idle in seconds
# (optional; default: 1140)
SESSION_TIMEOUT_SECONDS=1140
# how long before idle to emit the onPrompt event in seconds
# (optional; default: 300)
SESSION_TIMEOUT_PROMPT_SECONDS=300
```

frontend/app/root.tsx
```diff
@@ -38,6 +39,8 @@ export async function loader({ request }: LoaderFunctionArgs) {
         buildVersion: '0.0.0-00000000-0000',
       },
       env: publicEnv,
+      sessionTimeoutPrompt: privateEnv.SESSION_TIMEOUT_PROMPT_SECONDS,
+      sessionTimeout: privateEnv.SESSION_TIMEOUT_SECONDS,
```
nicholas-ly, line 43: These two values are going to be exposed publicly when the `<SessionTimeout />` renders. I feel there's no need to create a new `privateEnv` within the `loader()` to pass it to the `<SessionTimeout/>`.

We could alternatively take advantage of `env-utils`' `getClientEnv()` instead. For example:
https://github.com/DTS-STN/canadian-dental-care-plan/blob/aabe531e546f6adbccbb515a34b3e2037edc0327/frontend/app/components/language-switcher.tsx#L22

gregory-j-baker (dismissed):

frontend/app/root.tsx
gregory-j-baker: The `getClientEnv()` function is most useful when you need to pass config to a non-route component, but don't want to pass the value down through a bunch of nested components to get there.

In cases where we're rendering a component directly in a route, I think using `.getEnv()` inside a loader to pass that config to the route, and then the target component is okay.

One advantage to using a `.getClientEnv()` call inside a route would be to avoid blocking rendering while the loader data is being fetched.. but since we have to do an auth check for every route, this becomes moot.

nicholas-ly (approved):

nicholas-ly (approved):

PR #479, opened 2024-02-22
AB#2934 -- Added maxAge memo config to lookup service

AdamP-HRSDC-RHDCC:
### Description
Added environment variable configurable maxAge config to calls in lookup service. 
maxAge is the cache expiry time in milliseconds before the cache will expire (undefined for infinity)

[AB#2934](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/2934)

N/A

### Checklist
- [n/a] I have added/updated tests that prove my fix is effective or that my feature works
- [x] I have checked that all e2e tests pass by running `npm run test:e2e` (passed but is also timing out and failing)

### Test Instructions
Run and confirm no change in behavior on language and address pages

### Additional Notes
Used nested moize functions to keep singleton behavior


frontend/.env.example
```diff
@@ -120,4 +120,10 @@ MOCK_AUTH_ALLOWED_REDIRECTS=http://localhost:3000/auth/callback/raoidc
...
 # Community where the documents were generated from, and user belongs. This value is linked to the specific vault database for the given community.
-CCT_VAULT_COMMUNITY='SecurityReview'
\ No newline at end of file
+CCT_VAULT_COMMUNITY='SecurityReview'
+
+# Memoization: amount of time in milliseconds before the cache will expire (undefined for infinity)
+MEMO_MAXAGE_ALLPREFERREDLANGUAGES=0
+MEMO_MAXAGE_PREFERREDLANGUAGE=0
+MEMO_MAXAGE_ALLCOUNTRIES=0
+MEMO_MAXAGE_ALLREGIONS=0
```
gregory-j-baker, line 129: My suggestion here would be to add a time unit..
And it might be better to name the configs something a little more intuitive. (for example, `MEMO_MAXAGE` might not be as intuitive as `*_CACHE_TTL_SECONDS`)

```
LOOKUP_SVC_PREFERRED_LANGUAGES_CACHE_SECONDS
```

🤷

frontend/app/utils/env.server.ts
```diff
@@ -87,6 +87,12 @@ const serverEnv = z.object({
 
   // CCT get PDF settings
   CCT_VAULT_COMMUNITY: z.string().default('community_default'),
+
+  //Memo cache max age
+  MEMO_MAXAGE_ALLPREFERREDLANGUAGES: z.coerce.number().optional(),
+  MEMO_MAXAGE_PREFERREDLANGUAGE: z.coerce.number().optional(),
+  MEMO_MAXAGE_ALLCOUNTRIES: z.coerce.number().optional(),
+  MEMO_MAXAGE_ALLREGIONS: z.coerce.number().optional(),
```
gregory-j-baker, line 95: I think you should provide sane defaults here. If we forget to set these config values in production, we're going to hammer the downstream service.

gregory-j-baker (changes requested):

frontend/.env.example
AdamP-HRSDC-RHDCC: Updated naming schema to use LOOKUP_SVC_******_CACHE_TTL_MILLISECONDS

frontend/app/utils/env.server.ts
AdamP-HRSDC-RHDCC: removed optional and added default of 1 hr (still includes undefined as response type)

frontend/app/utils/env.server.ts
```diff
@@ -87,6 +87,12 @@ const serverEnv = z.object({
 
   // CCT get PDF settings
   CCT_VAULT_COMMUNITY: z.string().default('community_default'),
+
+  //Memo cache max age
+  LOOKUP_SVC_ALLPREFERREDLANGUAGES_CACHE_TTL_MILLISECONDS: z.coerce.number().default(3600000),
+  LOOKUP_SVC_PREFERREDLANGUAGE_CACHE_TTL_MILLISECONDS: z.coerce.number().default(3600000),
+  LOOKUP_SVC_ALLCOUNTRIES_CACHE_TTL_MILLISECONDS: z.coerce.number().default(3600000),
+  LOOKUP_SVC_ALLREGIONS_CACHE_TTL_MILLISECONDS: z.coerce.number().default(3600000),
```
gregory-j-baker, line 95: **Nitpick:** (maybe change in a future PR)

Consider using `z.coerce.number().default(60 * 60 * 1000)` to make it a little more obvious that this is 1h.

gregory-j-baker (approved):

PR #492, opened 2024-02-23
Ignore `/api/readyz` request logging in express

gregory-j-baker:
### Description

Avoid logging `/api/readyz`. Kubernetes hits this endpoint so frequently that it is filling up our logs.


sebastien-comeau:
Can you change the log category level with [winston](https://github.com/winstonjs/winston)?

gregory-j-baker:
> Can you change the log category level with [winston](https://github.com/winstonjs/winston)?

It's definitely *possible*. Our logging solution currently doesn't let us configure the logging level *per category*, though. Adding that feature would have to be a future enhancement.

sebastien-comeau (approved):

PR #509, opened 2024-02-26
change e2e folder structure

Fbasham:
Change the folder structure of end-to-end tests to match remix-flat-routes style like how unit tests and the application itself are currently structured.


sebastien-comeau (approved):

Fbasham:
> Can we match the exact structure and names of the unit tests and routes directory?
> 
> **e2e structure in this branch:** image
> 
> **Unit tests structure:** image

Oh yeah..derp.  I'm totally on auto-pilot right now.

sebastien-comeau:
> Can we match the exact structure and names of the unit tests and routes directory?
> 
> **e2e structure in this branch:** image
> 
> **Unit tests structure:** image

Well, it doesn't have to be similar to unit tests since the inted of e2e is to test routes only.

nicholas-ly:
> Well, it doesn't have to be similar to unit tests since the inted of e2e is to test routes only.

Agreed though the e2e tests are (coincidentally?) named similarly to unit tests and its routes, so why not keep the consistency

sebastien-comeau:
> > Well, it doesn't have to be similar to unit tests since the inted of e2e is to test routes only.
> 
> Agreed though the e2e tests are (coincidentally?) named similarly to unit tests and its routes, so why not keep the consistency

My bad, I thought `_app+` was a mistake.

PR #516, opened 2024-02-27
Application Type Form Added

CanadianPhilipBarkow:
Application type form now added, including radio button with custom validation message to allow for translation. 

[AB#2931](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/2931)

Screenshot from 2024-02-27 06-59-50

### Checklist

### Test Instructions

1. Run the application.
2. Navigate to the index page of the application.
3. Select the new application link
4. Click the Start application button
5. Observe the application type form
6. Click the continue button
7. Observe the error message
8. Toggle language
9. Click the continue button
10. Observe the error message

### Additional Notes

Other radio buttons should also be updated to allow for translation.

### DIFF
--- a/frontend/app/routes/_app+/new-application+/a2.tsx
+++ b/frontend/app/routes/_app+/new-application+/a2.tsx
@@ -1,0 +1,83 @@
+import { json, redirect } from '@remix-run/node';
+import type { ActionFunctionArgs, LoaderFunctionArgs } from '@remix-run/node';
+import { Form, useLoaderData } from '@remix-run/react';
+
+import { useTranslation } from 'react-i18next';
+
+import { Button, ButtonLink } from '~/components/buttons';
+import { InputRadios } from '~/components/input-radios';
+import { getRaoidcService } from '~/services/raoidc-service.server';
+import { getSessionService } from '~/services/session-service.server';
+import { getTypedI18nNamespaces } from '~/utils/locale-utils';
+
+const i18nNamespaces = getTypedI18nNamespaces('new-application');
+
+export const handle = {
+  breadcrumbs: [{ labelI18nKey: 'new-application:a2.breadcrumbs.new-application' }, { labelI18nKey: 'new-application:a2.breadcrumbs.a2' }],
+  i18nNamespaces,
+  pageIdentifier: 'CDCP-0016',
+  pageTitleI18nKey: 'new-application:a2.page-title',
+};
+
+export async function loader({ request }: LoaderFunctionArgs) {
+  const raoidcService = await getRaoidcService();
+  await raoidcService.handleSessionValidation(request);
+
+  const sessionService = await getSessionService();
+  const session = await sessionService.getSession(request);
+  const selectedType = session.get('selectedType');
+
+  return json({ selectedType });
+}
+
+export async function action({ request }: ActionFunctionArgs) {
+  const raoidcService = await getRaoidcService();
+  await raoidcService.handleSessionValidation(request);
+
+  const sessionService = await getSessionService();
+  const session = await sessionService.getSession(request);
+
+  const formDataRadio = Object.fromEntries(await request.formData());
+  const selectedType = formDataRadio.selectedType === 'self';
+  session.set('selectedType', selectedType);
+
+  return redirect('/new-application/a3', {
+    headers: {
+      'Set-Cookie': await sessionService.commitSession(session),
+    },
+  });
+}
+
+export default function A2() {
+  useLoaderData<typeof loader>();
+  const { t } = useTranslation(i18nNamespaces);
+  const updateValidationMessage = () => {
+    (document.getElementById('input-radio-selected-type-option-0') as HTMLInputElement).setCustomValidity(t('new-application:a2.required'));
+  };
+
+  return (
+    <>
+      <p className="mb-4 text-red-600">{t('gcweb:asterisk-indicates-required-field')}</p>
+      <Form method="post">
+        <InputRadios
+          id="selected-type"
+          name="selectedType"
+          legend={t('new-application:a2.choose-type')}
+          options={[
+            { value: 'self', children: t('new-application:a2.self') },
+            { value: 'else', children: t('new-application:a2.else') },
+          ]}
+          required
+        />
+        <div className="flex flex-wrap items-center gap-3">
+          <ButtonLink id="cancel-button" to="/new-application/">
+            {t('new-application:a2.back')}
+          </ButtonLink>
+          <Button id="confirm-button" variant="primary" onClick={updateValidationMessage}>
+            {t('new-application:a2.continue')}
+          </Button>
+        </div>
+      </Form>
+    </>
+  );
+}
--- a/frontend/public/locales/en/new-application.json
+++ b/frontend/public/locales/en/new-application.json
@@ -30,15 +30,15 @@
       "button": "Start application"
     }
   },
-  "a2": {
+  "application-type": {
     "page-title": "Type of application",
     "breadcrumbs": {
       "new-application": "New Application",
-      "a2": "Type of Application"
+      "application-type": "Type of Application"
     },
     "choose-type": "I am",
     "self": "Applying for myself",
-    "else": "Applying on behalf of someone else",
+    "other": "Applying on behalf of someone else",
     "back": "Back",
     "continue": "Continue",
     "required": "Please select one of these options"
--- a/frontend/public/locales/en/new-application.json
+++ b/frontend/public/locales/en/new-application.json
@@ -29,5 +29,18 @@
       "paragraph-one": "By starting this application, you consent to the terms of the above privacy notice statement.",
       "button": "Start application"
     }
+  },
+  "a2": {
+    "page-title": "Type of application",
+    "breadcrumbs": {
+      "new-application": "New Application",
+      "a2": "Type of Application"
+    },
+    "choose-type": "I am",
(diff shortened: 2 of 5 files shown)


frontend/app/routes/_app+/new-application+/a2.tsx
gregory-j-baker, line 1: Is there any particular reason you chose `a2` as the name for this route? Why not something a little more descriptive?

frontend/app/routes/_app+/new-application+/a2.tsx
```diff
@@ -0,0 +1,83 @@
...
+    },
+  });
+}
+
+export default function A2() {
+  useLoaderData<typeof loader>();
+  const { t } = useTranslation(i18nNamespaces);
+  const updateValidationMessage = () => {
+    (document.getElementById('input-radio-selected-type-option-0') as HTMLInputElement).setCustomValidity(t('new-application:a2.required'));
+  };
```
gregory-j-baker, line 56: This is an interesting way to handle form validation, but it doesn't really follow the canada.ca way of doing it. I'm interested in what @sebastien-comeau and @nicholas-ly think about this.

Also.. this approach doesn't survive a language toggle, which is pretty important (in my opinion).

frontend/app/routes/_app+/new-application+/a2.tsx
```diff
@@ -0,0 +1,83 @@
...
+export async function action({ request }: ActionFunctionArgs) {
+  const raoidcService = await getRaoidcService();
+  await raoidcService.handleSessionValidation(request);
+
+  const sessionService = await getSessionService();
+  const session = await sessionService.getSession(request);
+
+  const formDataRadio = Object.fromEntries(await request.formData());
+  const selectedType = formDataRadio.selectedType === 'self';
+  session.set('selectedType', selectedType);
```
gregory-j-baker, line 42: Two things for this:
1. I don't think this needs to be stored server-side. Couldn't you just redirect to the correct next page depending on the selected value?
1. If it is required for some reason, the name of the variable doesn't really match the values.. `selectedType: true` and `selectedType: false`.

sebastien-comeau:
> In Remix the form is posted to the route action method. The validation occurs in the action and errors are returned if any are found then the react route component will react on those errors. https://github.com/DTS-STN/canadian-dental-care-plan/blob/main/frontend/app/routes/_app%2B/personal-information%2B/phone-number%2B/edit.tsx

https://remix.run/docs/en/main/guides/form-validation

frontend/app/routes/_app+/new-application+/a2.tsx
```diff
@@ -0,0 +1,83 @@
...
+        <InputRadios
+          id="selected-type"
+          name="selectedType"
+          legend={t('new-application:a2.choose-type')}
+          options={[
+            { value: 'self', children: t('new-application:a2.self') },
+            { value: 'else', children: t('new-application:a2.else') },
+          ]}
+          required
+        />
```
gregory-j-baker, line 71: **Nitpick:** I would name this input and its values something a little more descriptive. Perhaps `applyingFor: self` and `applyingFor: other`.

frontend/public/locales/en/new-application.json
```diff
@@ -29,5 +29,18 @@
       "paragraph-one": "By starting this application, you consent to the terms of the above privacy notice statement.",
       "button": "Start application"
     }
+  },
+  "a2": {
+    "page-title": "Type of application",
+    "breadcrumbs": {
+      "new-application": "New Application",
+      "a2": "Type of Application"
```
gregory-j-baker, line 37: **Nitpick:** `a2` could be named something a little more descriptive, in my opinion.

gregory-j-baker (changes requested):

frontend/app/routes/_app+/new-application+/a2.tsx
sebastien-comeau: Yup, TIL but it has some complexity since we are managing errors on the route's action. Even the GCWeb example doesn't override the default browser error messages on the client.
https://wet-boew.github.io/gcweb-compiled-demos/wetboew-demos/formvalid/formvalid-fr.html

frontend/app/routes/_app+/new-application+/a2.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  return redirect('/new-application/a3', {
+    headers: {
+      'Set-Cookie': await sessionService.commitSession(session),
+    },
+  });
+}
+
+export default function A2() {
+  useLoaderData<typeof loader>();
```
nicholas-ly, line 52: `useLoaderData()` isn't assigned to any variable and can be removed.

CanadianPhilipBarkow:
> In Remix the form is posted to the route action method. The validation occurs in the action and errors are returned if any are found then the react route component will react on those errors. https://github.com/DTS-STN/canadian-dental-care-plan/blob/main/frontend/app/routes/_app%2B/personal-information%2B/phone-number%2B/edit.tsx

While that is of course relevant to other parts of the form, the use of the 'required' property in the `inputradio` element triggers validation on the dom level.

Unless you mean that I should override the existing dom level validation entirely?

sebastien-comeau:
> > In Remix the form is posted to the route action method. The validation occurs in the action and errors are returned if any are found then the react route component will react on those errors. https://github.com/DTS-STN/canadian-dental-care-plan/blob/main/frontend/app/routes/_app%2B/personal-information%2B/phone-number%2B/edit.tsx
> 
> While that is of course relevant to other parts of the form, the use of the 'required' property in the `inputradio` element triggers validation on the dom level.
> 
> Unless you mean that I should override the existing dom level validation entirely?

The 'required' input attribute performs client-side validation as it's a browser feature. We don't intend to modify the default error messages provided by the browser. Primary form validation needs to occur within the form action in case the client disables JavaScript.

frontend/app/routes/_app+/new-application+/a2.tsx
CanadianPhilipBarkow: It came from the mockup, it has now been updated.

frontend/app/routes/_app+/new-application+/a2.tsx
CanadianPhilipBarkow: Admittedly, leaving the browser error message untranslated seems inconsistent to me, but if that's the standard... ¯\\_(ツ)_/¯ 

Now removed.

frontend/app/routes/_app+/new-application+/a2.tsx
CanadianPhilipBarkow: 1. I misread the mockup, you're right that doesn't need to be stored server side.
2. I've updated it.

frontend/app/routes/_app+/new-application+/a2.tsx
CanadianPhilipBarkow: Updated.

frontend/public/locales/en/new-application.json
CanadianPhilipBarkow: Updated.

frontend/app/routes/_app+/new-application+/a2.tsx
CanadianPhilipBarkow: Removed, thanks.

frontend/app/routes/_app+/new-application+/a2.tsx
gregory-j-baker: When handling the form validation in the Remix route, the browser error message isn't triggered, so there's no need to worry about its translation. Look at how error handling is done in <https://github.com/DTS-STN/canadian-dental-care-plan/blob/main/frontend/app/routes/_app%2B/personal-information%2B/phone-number%2B/edit.tsx>

image

gregory-j-baker:
> While that is of course relevant to other parts of the form, the use of the 'required' property in the `inputradio` element triggers validation on the dom level.

You're correct. This is something I haven't really paid much attention to before.

image

Maybe we should consider translating these.. or, we don't use the browser's `required` attribute on inputs and instead rely on our own custom validation to catch missing values. Since we have to do this for other validation rules, I feel like this might be the better solution.

gregory-j-baker (approved):

frontend/app/routes/_app+/new-application+/a2.tsx
CanadianPhilipBarkow: Screenshot from 2024-02-28 07-03-24

This looks like the browser error message is being triggered, I don't see that error message anywhere in the project.

gregory-j-baker: You're right. I think we should fix this in all of our forms. It's not the best user experience.

sebastien-comeau:
> > While that is of course relevant to other parts of the form, the use of the 'required' property in the `inputradio` element triggers validation on the dom level.
> 
> You're correct. This is something I haven't really paid much attention to before.
> 
> image
> 
> Maybe we should consider translating these.. or, we don't use the browser's `required` attribute on inputs and instead rely on our own custom validation to catch missing values. Since we have to do this for other validation rules, I feel like this might be the better solution.

AFAIk, `required` attr. is reqired for a11y. I think there's a way to disable the form client validation with an attr.

sebastien-comeau:
@CanadianPhilipBarkow @gregory-j-baker 
HTML <form> novalidate Attribute
https://www.w3schools.com/tags/att_form_novalidate.asp

PR #589, opened 2024-03-04
AB#2997 refactor date of birth error handling

Fbasham:
### Description
Refactor the date of birth on the spousal information form page to using a single refinement on the separate date fields (month, day, year) to validate whether the inputted date is actually valid (handles leap years, out of range values,etc). Also add unit tests.


Fbasham:
> I think you should copy over and refactor the `DateSelectField` from PASC since it's already tested by a11y. https://github.com/DTS-STN/passport-status/blob/main/src/components/DateSelectField.tsx

So I looked into this and I'm not sure it makes a lot of sense at this current time to do that for a few reasons:
1. the figma mockups don't have the day/year being `<select />` elements
2. the complexity of those 2 components (DataSelect and DateSelectField) doesn't seem necessary in this PR (I can't see anywhere else we need to specifically have this component)
3. this PR really only addresses one small aspect of the date component (i.e. I was really only aiming to address the error handing)  Before this PR, there were separate error messages for each field, and this PR consolidates that error handling a single validation

I propose that if we really need a separate component for dates, then we can look into bringing it in.  Happy to listen to your reasoning though as well and address in a separate PR.

sebastien-comeau:
> So I looked into this and I'm not sure it makes a lot of sense at this current time to do that for a few reasons:
> 
> 1. the figma mockups don't have the day/year being `<select />` elements
> 2. the complexity of those 2 components (DataSelect and DateSelectField) doesn't seem necessary in this PR (I can't see anywhere else we need to specifically have this component)
> 3. this PR really only addresses one small aspect of the date component (i.e. I was really only aiming to address the error handing)  Before this PR, there were separate error messages for each field, and this PR consolidates that error handling a single validation
> 
> I propose that if we really need a separate component for dates, then we can look into bringing it in. Happy to listen to your reasoning though as well and address in a separate PR.

I disagree with the figma mockup. It should be 3 selects and the format should be year, month and day as it is in other GoC forms. Alos the [Canada's full date format](https://en.wikipedia.org/wiki/Date_and_time_notation_in_Canada) is not **month, day and year** in French but **day, month and year** unless it's all-numeric format (yyyy-mm-dd) which is the same in both languages. Can address this later if necessary.

I have a different perspective on the figma mockup. I suggest having three select options, and the order should be **year, month, and day**, aligning with the format used in other Government of Canada forms. Additionally, it's worth noting that [Canada's full date format](https://en.wikipedia.org/wiki/Date_and_time_notation_in_Canada) in French is not **month, day, and year** but rather **day, month, and year**, unless it's in an all-numeric format (yyyy-mm-dd), which remains consistent in both languages. We can tackle this issue later if needed.

frontend/app/routes/_public+/apply+/$id+/partner-information.tsx
```diff
@@ -133,14 +138,19 @@ export default function ApplyFlowApplicationInformation() {
...
-            <InputSelect id="month" label={t('partner-information.month')} options={monthOptions} name="month" errorMessage={errorMessages.month} />
-            <InputField id="day" label={t('partner-information.day')} name="day" type="number" min={1} max={31} errorMessage={errorMessages.day} />
-            <InputField id="year" label={t('partner-information.year')} name="year" type="number" min={1900} errorMessage={errorMessages.year} />
+            <InputSelect id="month" label={t('partner-information.month')} options={monthOptions} name="month" />
+            <InputField id="day" label={t('partner-information.day')} name="day" type="number" min={1} max={31} />
+            <InputField id="year" label={t('partner-information.year')} name="year" type="number" min={1900} />
           </div>
         </fieldset>
+        {errorMessages.dateOfBirth && (
+          <InputError id="date-of-birth-error" className="mt-2">
```
sebastien-comeau, line 150: To ensure accessibility, the ID should be designated as "aria-errormessage" for the corresponding inputs, with the inputs themselves having "aria-invalid" set to true.

PR #607, opened 2024-03-07
Make apply form multi-tab/multi-window safe

gregory-j-baker:
### Description

This PR addresses data clobbering issues that could occur when users fill out the application form across multiple tabs.

Here's what it does:

- **prevents data clobbering:** When a user opens the application form in a new tab (by middle-clicking, copying/pasting the URL, etc.), they'll be redirected back to the beginning of the form. This ensures each tab holds independent data and prevents accidental overwriting from other tabs.
- **stops accidental resubmissions:** Users won't be able to use the browser's back button to revisit and resubmit a previously submitted form. This helps avoid duplicate submissions and data inconsistencies.

- [AB#2973](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/2973)

### Checklist

### Test Instructions

- navigate to <http://localhost:3000/apply> and begin the application flow
- on any page after the initial start page, open the form in another tab (middle click, copy/paste URL, etc)
- verify that you are bounced back to the application form's start

### DIFF
--- a/frontend/app/routes/_public+/apply+/_index.tsx
+++ b/frontend/app/routes/_public+/apply+/_index.tsx
@@ -65,6 +65,8 @@ export default function ApplyIndex() {
 
       captchaRef.current.resetCaptcha();
     }
+
+    sessionStorage.setItem('flow.state', 'active');
   }
 
   return (
--- a/frontend/app/routes/_public+/apply+/$id+/_route.tsx
+++ b/frontend/app/routes/_public+/apply+/$id+/_route.tsx
@@ -1,0 +1,19 @@
+import { useEffect } from 'react';
+
+import { Outlet, useNavigate } from '@remix-run/react';
+
+/**
+ * The parent route of all /apply/{id}/* routes, used to
+ * redirect to /apply if the flow has not yet been initialized.
+ */
+export default function Route() {
+  const navigate = useNavigate();
+
+  useEffect(() => {
+    // redirect to start if the flow has not yet been initialized
+    const flowState = sessionStorage.getItem('flow.state');
+    if (flowState !== 'active') navigate('/apply');
+  }, [navigate]);
+
+  return <Outlet />;
+}
--- a/frontend/app/routes/_public+/apply+/$id+/application-delegate.tsx
+++ b/frontend/app/routes/_public+/apply+/$id+/application-delegate.tsx
@@ -51,7 +51,7 @@ export default function ApplyFlowApplicationDelegate() {
           {t('application-delegate.back-btn')}
           <FontAwesomeIcon icon={faChevronLeft} className="pl-2" />
         </ButtonLink>
-        <ButtonLink type="submit" variant="primary" to="/">
+        <ButtonLink type="submit" variant="primary" to="/" onClick={() => sessionStorage.removeItem('flow.state')}>
           {t('application-delegate.return-btn')}
         </ButtonLink>
       </div>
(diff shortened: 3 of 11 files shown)


frontend/app/routes/_public+/apply+/$id+/applicant-information.tsx
```diff
@@ -60,6 +60,14 @@ export async function action({ request, params }: ActionFunctionArgs) {
...
 
 export default function ApplyFlowApplicationInformation() {
+  const navigate = useNavigate();
+
+  useEffect(() => {
+    // redirect to start if the flow has not yet been initialized
+    const flowState = sessionStorage.getItem('flow.state');
+    if (flowState !== 'active') navigate('/apply');
+  }, [navigate]);
+
```
sebastien-comeau, line 70: Is is possible to have this code in his own component?

gregory-j-baker: Yeah. I can work on that. Are you okay with the general approach, though?

- set state in starting route
- check state in non-starting routes
- reset state in end route

sebastien-comeau: Idk, if you go with `_layout.tsx` option... check if it render on location change.

sebastien-comeau: I think if you go with `_layout.tsx` and `useEffect` it will be good to use `location` as a `useEffect` dependency.
https://remix.run/docs/en/main/hooks/use-location#uselocation

gregory-j-baker: The problem with using `_layout.tsx` is that it will fire indiscriminately for all child routes.. but the behavior of the starting route is not the same as the behavior of the non-starting route.

And I don't really want to get into checking the URL to see if it's `=== '/apply'` because that's kinda gross and fragile.

sebastien-comeau: `_layout.tsx` will be inside `$id+` directory

gregory-j-baker: I added a `routes/_public+/apply+/$id+/_route.tsx` file that contains the `useEffect()`. Please re-review.

frontend/app/routes/_public+/apply+/_index.tsx
```diff
@@ -65,6 +65,8 @@ export default function ApplyIndex() {
 
       captchaRef.current.resetCaptcha();
     }
+
+    sessionStorage.setItem('flow.state', 'active');
```
sebastien-comeau, line 69: I think the `sessionStorage.setItem` and navigate should happened after the action was successful.

`action` should return `json` with the generated `id`.
```
return json({ id }, sessionResponseInit);
```

then react when `useActionData` have an `id`.  Set the sessionStorage, then redirect client-side.

gregory-j-baker: I'll give it a shot.

gregory-j-baker: I've been thinking about this a little bit and I don't like the idea because it violates the [Post/Redirect/Get](https://en.wikipedia.org/wiki/Post/Redirect/Get) pattern that we use to avoid the back-button-post problem.

Also, it would completely break non-javascript clients. I acknowledge that we decided to require javascript, but I do think we should try to maintain some functionality for clients who happen to have javascript disabled.

sebastien-comeau (approved):

PR #659, opened 2024-03-12
Adding state into review-info page.

nibivi77:
### Description
There are some fields that are not in `state`. Left them as hard-coded string, like `firstnames`

[AB#3001](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/3001)

### Checklist

### Test Instructions
Try to go through the apply flow (There are many flow bugs)
You will have type in `/apply/{id}/review-information` since the flow is not redirecting to the correct page.

### Additional Notes
I did not want to add too much logic in the `loader` since there are many ways to handle the state. I wanted to hear the team out, see what would be better and voted by everyone.


frontend/app/routes/_public+/apply+/$id+/review-information.tsx
```diff
@@ -231,30 +241,29 @@ export default function ReviewInformation() {
       <h2 className="mt-8 text-2xl font-semibold ">{t('apply:review-information.dental-title')}</h2>
       <dl>
         <DescriptionListItem term={t('apply:review-information.dental-private-title')}>
-          {dentalInsurance.private.length > 0
-            ? dentalInsurance.private.map((benefit) => {
-                return benefit;
-              })
-            : t('apply:review-information.dental-has-no-access')}
+          {dentalInsurance === 'yes' ? t('apply:review-information.dental-insurance-has-access') + dentalInsurance : t('apply:review-information.dental-no-access')}
```
gregory-j-baker, line 244: Move the `yes` string into the translation string. Otherwise, you could end up rendering:

```
A accès à une assurance ou une couverture dentaire : yes
```

Also if the value of `dentalInsurance` changes for some reason, then the page won't render as expected.

frontend/app/routes/_public+/apply+/$id+/review-information.tsx
```diff
@@ -231,30 +241,29 @@ export default function ReviewInformation() {
...
               {t('apply:review-information.dental-private-change')}
             </InlineLink>
           </p>
         </DescriptionListItem>
         <DescriptionListItem term={t('apply:review-information.dental-public-title')}>
-          {dentalInsurance.public.length > 0 ? (
+          {dentalBenefit.federalBenefit.access === 'yes' ? (
+            <div>
+              <span>{t('apply:review-information.dental-has-access')}</span>
+              <ul className="ml-6 list-disc">{dentalBenefit.federalBenefit.benefit}</ul>
```
gregory-j-baker, line 255: Is this supposed to be

```
<ul className="ml-6 list-disc"> 
  <li>{dentalBenefit.federalBenefit.benefit}</li>
</ul>
```

maybe? It's hard to tell what's in that variable.

frontend/app/routes/_public+/apply+/$id+/review-information.tsx
```diff
@@ -231,30 +241,29 @@ export default function ReviewInformation() {
...
+          {dentalBenefit.provTerrBenefit.access === 'yes' ? (
             <div>
               <span>{t('apply:review-information.dental-has-access')}</span>
-              <ul className="ml-6 list-disc">
-                {dentalInsurance.public.map((benefit) => {
-                  // eslint-disable-next-line react/jsx-key
-                  return <li key={benefit.id}>{benefit.benefitEn}</li>;
-                })}
-              </ul>
+              <ul className="ml-6 list-disc">{dentalBenefit.provTerrBenefit.benefit}</ul>
```
gregory-j-baker, line 263: Is this supposed to be wrapped in an `<li>` element, like the one above?

gregory-j-baker (changes requested):

frontend/app/routes/_public+/apply+/$id+/review-information.tsx
```diff
@@ -33,58 +34,66 @@ export const meta: MetaFunction<typeof loader> = mergeMeta(({ data }) => {
...
-    city: 'Lake Granvillestead',
-    province: 'PE',
-    postalCode: 'P1L 1C5',
-    country: 'CAN',
+    address: state.personalInformation?.mailingAddress ?? 'Unknown',
+    appartment: state.personalInformation?.mailingApartment ?? 'Unknown',
+    city: state.personalInformation?.mailingCity ?? 'Unknown',
+    province: state.personalInformation?.mailingProvince ?? 'Unknown',
+    postalCode: state.personalInformation?.mailingPostalCode ?? 'Unknown',
+    country: state.personalInformation?.mailingCountry ?? 'Unknown',
```
nicholas-ly, line 70: I dislike having all these default values for what I assume are mandatory fields for the application. I wonder if we should have a different validation schema for the entire application rather than re-using the flow state schema, which has every page in the flow as optional.

cc @gregory-j-baker @sebastien-comeau

gregory-j-baker: I think the validation should be moved into their respective pages, and exported from there.

The final step that will store the data via an API call should then validate the data as a whole before submitting. This can be done by simply setting a `validated: true` flag in the data for a specific step, or by re-running the data through the exported validators.

As for the defaults here, I just kind of assumed that this is unfinished placeholder code (is it not?).

gregory-j-baker: Sorry.. to be clear, my comment about moving the validation into each respective .tsx file is not for this PR.. so don't worry too much about that.

nicholas-ly: Resolving for future PR

nicholas-ly (approved):

PR #677, opened 2024-03-14
tailwind print styles

Fbasham:
### Description
Add tailwind print styles so that we can hide elements when printing and also render hyperlinks in a readable format.

AB#3062

image


gregory-j-baker (approved):
I feel like this bit of text should probably be hidden, too:

image

> you can use this button to print

WHAT BUTTON?!

PR #690, opened 2024-03-15
add spinner on form submit

Fbasham:
### Description
Add loading spinner when apply form is 'submitting'.  NOTE: I've only added this to the forward navigations and not the back navigations.  Do we need it on back navigations?  Open to suggestions.  Keep in mind, the 'back buttons' are actually `ButtonLinks`, so semantically similar to anchors. Does a spinner go on an anchor usually?  I don't have an answer, but intuition says 'no'.

AB#3072


frontend/app/routes/$lang+/_public+/apply+/$id+/applicant-information.tsx
```diff
@@ -157,9 +158,9 @@ export default function ApplyFlowApplicationInformation() {
             <FontAwesomeIcon icon={faChevronLeft} className="me-3 block size-4" />
             {t('applicant-information.back-btn')}
           </ButtonLink>
-          <Button variant="primary" id="continue-button">
+          <Button variant="primary" id="continue-button" disabled={navigation.state === 'submitting'}>
             {t('applicant-information.continue-btn')}
-            <FontAwesomeIcon icon={faChevronRight} className="ms-3 block size-4" />
+            <FontAwesomeIcon icon={navigation.state === 'submitting' ? faSpinner : faChevronRight} className={`ms-3 block size-4 ${navigation.state === 'submitting' ? 'animate-spin' : ''}`} />
```
nicholas-ly, line 163: **Nitpick:** To remove the use of ternary operator (can be applied to other files in this PR):

```suggestion
            <FontAwesomeIcon icon={navigation.state === 'submitting' ? faSpinner : faChevronRight} className={`ms-3 block size-4 ${navigation.state === 'submitting' && 'animate-spin'}`} />
```

gregory-j-baker, line 163: For this I think you should use

```
className={cn('ms-3 block size-4', navigation.state !== 'idle' && 'animate-spin')}
```

since the `cn()` function was designed for this specific purpose

nicholas-ly (approved):
I agree that we shouldn't add the spinner to those "Back" buttons. 

Disabling these "Back" buttons when the form is submitting on the other hand... :thinking:

gregory-j-baker (changes requested):

frontend/app/routes/$lang+/_public+/apply+/$id+/applicant-information.tsx
gregory-j-baker: If you do this and `navigation.state !== 'submitting`, you will end up with `class="... false"` in the output.

The `cn()` function will remove that `false`.

gregory-j-baker (approved):

PR #723, opened 2024-03-19
AB3090 move applicant info schema

Fbasham:
### Description
Move applicant info schema to its respective route

[AB#3090](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/3090)

### DIFF
--- a/frontend/app/routes/$lang+/_public+/apply+/$id+/applicant-information.tsx
+++ b/frontend/app/routes/$lang+/_public+/apply+/$id+/applicant-information.tsx
@@ -21,10 +21,15 @@ import { getFixedT, redirectWithLocale } from '~/utils/locale-utils.server';
 import { mergeMeta } from '~/utils/meta-utils';
 import { RouteHandleData } from '~/utils/route-utils';
 import { getTitleMetaTags } from '~/utils/seo-utils';
-import { formatSin } from '~/utils/sin-utils';
+import { formatSin, isValidSin } from '~/utils/sin-utils';
 import { cn } from '~/utils/tw-utils';
 
-export const applyIdParamSchema = z.string().uuid();
+export type ApplicantInformationState = {
+  socialInsuranceNumber: string;
+  firstName: string;
+  lastName: string;
+  maritalStatus: string;
+};
 
 export const handle = {
   i18nNamespaces: getTypedI18nNamespaces('apply', 'gcweb'),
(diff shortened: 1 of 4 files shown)


frontend/app/routes/$lang+/_public+/apply+/$id+/applicant-information.tsx
```diff
@@ -50,14 +55,25 @@ export async function loader({ request, params }: LoaderFunctionArgs) {
 export async function action({ request, params }: ActionFunctionArgs) {
   const applyFlow = getApplyFlow();
   const { id } = await applyFlow.loadState({ request, params });
+  const t = await getFixedT(request, handle.i18nNamespaces);
+
+  /**
+   * Schema applicant information.
+   */
+  const applicantInformationSchema: z.ZodType<ApplicantInformationState> = z.object({
```
sebastien-comeau, line 63: Use `z.string().trim();` with strings to remove whitespaces before calling `.min(1)`.

frontend/app/routes/$lang+/_public+/apply+/$id+/applicant-information.tsx
```diff
@@ -50,14 +55,25 @@ export async function loader({ request, params }: LoaderFunctionArgs) {
...
+    socialInsuranceNumber: z.string().refine(isValidSin, { message: t('apply:applicant-information.error-message.valid-sin') }),
+    firstName: z.string().min(1, { message: t('apply:applicant-information.error-message.first-name') }),
+    lastName: z.string().min(1, { message: t('apply:applicant-information.error-message.last-name') }),
+    maritalStatus: z.string({ required_error: t('apply:applicant-information.error-message.marital-status') }),
+  });
 
   const formData = Object.fromEntries(await request.formData());
-  const parsedDataResult = applyFlow.applicantInformationSchema.safeParse(formData);
+  const parsedDataResult = applicantInformationSchema.safeParse(formData);
```
sebastien-comeau, line 71: Use [FormData: get() method](https://developer.mozilla.org/en-US/docs/Web/API/FormData/get) instead of using `Object.fromEntries`. We've started to use `.get()` to have more control on the data object that will be validated. As an example if you want to have all the checked values from a list of checboxes then we will use `.getAll()` that returns an arrays of values.

Here's an example how I did it.
https://github.com/DTS-STN/canadian-dental-care-plan/commit/5a6626edeb42a2713331650dbeb6e0eaaf3ae645#diff-34f5d4f6139e9c7445d3f867cb14241817fec4f7c0044f3026b88bdc604a9d22R67

frontend/app/routes/$lang+/_public+/apply+/$id+/applicant-information.tsx
sebastien-comeau: I've noticed that we can only return the message string as second argument instead of `{ message: "Error here!" }` object.

sebastien-comeau (changes requested):

frontend/app/routes/$lang+/_public+/apply+/$id+/applicant-information.tsx
```diff
@@ -50,14 +55,40 @@ export async function loader({ request, params }: LoaderFunctionArgs) {
...
+      .refine(isValidSin, { message: t('apply:applicant-information.error-message.valid-sin') }),
+    firstName: z
+      .string()
+      .trim()
+      .min(1, { message: t('apply:applicant-information.error-message.first-name') }),
+    lastName: z
+      .string()
+      .trim()
+      .min(1, { message: t('apply:applicant-information.error-message.last-name') }),
+    maritalStatus: z.string({ required_error: t('apply:applicant-information.error-message.marital-status') }),
```
sebastien-comeau, line 76: Here's my suggestion since the error message should be the same with the following scenarios:

- If the field is not defined (`undefined`), it displays the `required_error` message.
- If the field type is not `string`, specifically triggered when `null` is returned due to no marital statuses being selected, it shows the `invalid_type_error` message.
- If the selected marital status value is empty, it presents the `min` message.

```suggestion
    maritalStatus: z
      .string({ errorMap: () => ({ message: t('apply:applicant-information.error-message.marital-status') }) })
      .trim()
      .min(1, t('apply:applicant-information.error-message.marital-status')),
```

ATM, here's the error message that appears when the user hasn't selected anything:

image

sebastien-comeau (changes requested):

frontend/app/routes/$lang+/_public+/apply+/$id+/applicant-information.tsx
sebastien-comeau: To only leverage on `.min()`, I highly suggest your convert the returned value from `FormData.get()` to the corresponding type:

```tsx
const data = {
    socialInsuranceNumber: String(formData.get('socialInsuranceNumber') ?? ""),
    firstName: String(formData.get('firstName') ?? ""),
    lastName: String(formData.get('lastName') ?? ""),
    maritalStatus: String(formData.get('maritalStatus') ?? ""),
  };
```

**_Note:_** ` ?? ""` is added to fallback to an empty string when `null` is returned.

Fbasham: okay, fair points.

sebastien-comeau (approved):

sebastien-comeau (approved):

PR #761, opened 2024-03-21
AB#3088 Fix apply non existent routes

Fbasham:
### Description
Redirect user if they hit `/apply/$id`. 

[AB#3088](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/3088)

### DIFF
--- a/frontend/app/routes/$lang+/_public+/apply+/$id+/index.tsx
+++ b/frontend/app/routes/$lang+/_public+/apply+/$id+/index.tsx
@@ -1,0 +1,12 @@
+import { LoaderFunctionArgs } from '@remix-run/node';
+
+import { getApplyFlow } from '~/routes-flow/apply-flow';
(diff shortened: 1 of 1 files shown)


gregory-j-baker (commented):
I wonder if a better way to do this would be to create an `index.tsx` file with a loader that sends the user to the correct start page.

sebastien-comeau:
> I wonder if a better way to do this would be to create an `index.tsx` file with a loader that sends the user to the correct start page.

This was my first thought.

Fbasham:
> I wonder if a better way to do this would be to create an `index.tsx` file with a loader that sends the user to the correct start page.

True.  I think it's better to do it on the server than on redirection client side.  Good idea and I made the change.

sebastien-comeau (dismissed):

gregory-j-baker (approved):

nicholas-ly (approved):

PR #767, opened 2024-03-21
AB#3138 Remove partner information from state if Marital Status is not common in law and married

Fbasham:
### Description
Remove `partnerInformation` from state if and only if `applicantInformation` declares that their marital status is not married or common law. 

Before anybody asks, why not put the logic to remove parts of the state in the route action?  I did actually attempt that, but for whatever reason, the old state was being persisted even after a few attempts at saving the state with `partnerInformation` cleared.  Not sure if that's by design, so I went with a different approach here.

[AB#3138](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/3138)

### DIFF
--- a/frontend/app/routes-flow/apply-flow.ts
+++ b/frontend/app/routes-flow/apply-flow.ts
@@ -75,6 +75,7 @@ interface SaveStateArgs {
   params: Params;
   request: Request;
   state: ApplyState;
+  remove?: keyof ApplyState;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


frontend/app/routes-flow/apply-flow.ts
```diff
@@ -87,17 +87,22 @@ interface SaveStateArgs {
...
+  remove?: keyof ApplyState;
 }
 
 /**
  * Saves state.
  * @param args - The arguments.
  * @returns The Set-Cookie header to be used in the HTTP response.
  */
-async function saveState({ params, request, state }: SaveStateArgs) {
+async function saveState({ params, request, state, remove = undefined }: SaveStateArgs) {
```
gregory-j-baker, line 98: **Nitpick:** isn't the default value `undefined` already?

gregory-j-baker (approved):

gregory-j-baker:
If you enable auto merge (on future PRs) then I can rebase and let it merge for you when you're not around.

PR #781, opened 2024-03-22
AB#3136 Refine session for Unauthenticated space

Fbasham:
### Description
Looking for feedback on this one.  

From the ADO card:

> Add a timestamp in the session
> Once its submitted, add the current time
> On every page, if there is a timestamp and its within 15 min, redirect to the confirm page
> On every page, if there is a timestamp and its over 15 min, redirect to the apply page and kill session

Here's how this is working:
- when a user starts the apply flow, a new key is stored in session called `timeUpdated` with a value of the current timestamp
- when a user submits to the action in a route, the `timeUpdated` is set to the current timestamp
- if the difference between the current timestamp and the session `timeUpdated` is greater than 15 minutes, the session is killed and the user is redirected to `/apply`

There are some outstanding questions here though:
1. does there need to be a countdown timer to notify the user of how much time they have?  Or is a message at the start of the form in the instructions sufficient?  Accessibility will likely tell us both
2. Is the 15 minute limit to complete the form from start to finish, or can it be continuously reset as long as the user is active?  The ADO card is ambiguous here

[AB#3136](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/3136)

### DIFF
--- a/frontend/app/routes-flow/apply-flow.ts
+++ b/frontend/app/routes-flow/apply-flow.ts
@@ -68,6 +68,17 @@ async function loadState({ params, request, fallbackRedirectUrl = '/apply' }: Lo
   }
 
   const state = session.get(sessionName) as ApplyState;
+
+  if (Date.now() - session.get('timeUpdated') > 15 * 60 * 1000) {
+    session.unset(sessionName);
+    session.unset('timeUpdated');
+    throw redirectWithLocale(request, fallbackRedirectUrl, {
+      headers: {
+        'Set-Cookie': await sessionService.commitSession(session),
+      },
+    });
+  }
+
   return { id: id.data, state };
 }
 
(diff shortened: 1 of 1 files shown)


frontend/app/routes-flow/apply-flow.ts
```diff
@@ -68,6 +68,18 @@ async function loadState({ params, request, fallbackRedirectUrl = '/apply' }: Lo
   }
 
   const state = session.get(sessionName) as ApplyState;
+
+  if (Date.now() - session.get('timeUpdated') > 15 * 60 * 1000) {
+    session.unset(sessionName);
+    session.unset('timeUpdated');
+    throw redirectWithLocale(request, fallbackRedirectUrl, {
+      status: 302,
```
gregory-j-baker, line 76: `status: 302` is unnecessary here since the underlying `redirectWithLocale()` function calls remix's `redirect()` function.

gregory-j-baker (changes requested):

gregory-j-baker (approved):

PR #784, opened 2024-03-22
User Story 3050: More refactoring

carolinegeorge-gov:
### Description

Cleaned up personal information mocks (again).

[AB#3050](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/3050)

### Checklist


frontend/app/services/personal-information-service.server.ts
```diff
@@ -6,106 +6,113 @@ import { getLogger } from '~/utils/logging.server';
...
+          }),
+          PreferredIndicator: z.boolean(),
+        })
+        .optional(),
+      PersonSINIdentification: z.object({
+        IdentificationID: z.string(),
+      }),
+    }),
+  })
+  .optional();
```
nicholas-ly, line 69: Is there any reason why the entire API response object can be optional? We should either expect a response body represented as a personal info object back with a 200, or a 204 where we wouldn't parse the response body with this schema.

frontend/app/services/personal-information-service.server.ts
```diff
@@ -145,23 +152,26 @@ function createPersonalInformationService() {
...
+        ? personalInformationApiResponse.Client.ClientIdentification.filter((clientInfoDto) => clientInfoDto.IdentificationCategoryText === 'Applicant ID').at(0)?.IdentificationID
+        : '',
+      clientId: personalInformationApiResponse?.Client.ClientIdentification ? personalInformationApiResponse.Client.ClientIdentification.filter((clientInfoDto) => clientInfoDto.IdentificationCategoryText === 'Client Number').at(0)?.IdentificationID : '',
+      firstName: personalInformationApiResponse?.Client.PersonName ? personalInformationApiResponse.Client.PersonName.at(0)?.PersonGivenName?.at(0) : '',
+      lastName: personalInformationApiResponse?.Client.PersonName ? personalInformationApiResponse.Client.PersonName.at(0)?.PersonSurName?.at(0) : '',
+      emailAddress: personalInformationApiResponse?.Client.PersonContactInformation ? personalInformationApiResponse.Client.PersonContactInformation.EmailAddress.EmailAddressID : '',
+      phoneNumber: personalInformationApiResponse?.Client.PersonContactInformation ? personalInformationApiResponse.Client.PersonContactInformation.TelephoneNumber.FullTelephoneNumber : '',
```
nicholas-ly, line 167: Is it possible to remove the defaulted empty values for each of these fields and leave them as `undefined` instead? Our `personalInfoDtoSchema` indicates that these fields are optional.

We should be able to use optional chaining and do something like:

```suggestion
      phoneNumber: personalInformationApiResponse?.Client.PersonContactInformation?.TelephoneNumber.FullTelephoneNumber,
```

frontend/app/services/personal-information-service.server.ts
```diff
@@ -6,106 +6,113 @@ import { getLogger } from '~/utils/logging.server';
...
-  phoneNumber: z.string().optional(),
-  language: z
-    .object({
-      languageCode: z.string().optional(),
-      preferredIndicator: z.boolean().optional(),
-    })
-    .optional(),
-});
+  })
+  .optional();
```
nicholas-ly, line 108: Similar comment as the one made on `personalInformationApiResponseSchema`.

carolinegeorge-gov: I understand for personalInformationApiResponseSchema but if we were to add the call on the frontend, wouldn't React complain it can be null? Olivier had mentioned running into it when adding it.
image

nicholas-ly: `getPersonalInformation(..)` currently returns `null` if the response is a 204 though. 

I can't tell from that screenshot where that code comes from, but could it be you're doing a not-null assertion on a promise? Maybe we're missing an `await` since `getPersonalInformation` is currently `async`?

```typescript
const personalInformation = await getPersonalInformation(sin)
if (!personalInformation) { throw redirect(..); }
```

frontend/app/services/personal-information-service.server.ts
carolinegeorge-gov: I was doing it for another error but it looks like it wasn't needed. Fixed 102180defa666472647423336ecff5b8c2e4f0c2

frontend/app/services/personal-information-service.server.ts
carolinegeorge-gov: Removed: 102180defa666472647423336ecff5b8c2e4f0c2

frontend/app/services/personal-information-service.server.ts
carolinegeorge-gov: Removed: 102180defa666472647423336ecff5b8c2e4f0c2

nicholas-ly (approved):

PR #792, opened 2024-03-25
AB#2977 -- Trigger a challenge request only when the hCaptcha API loads

nicholas-ly:
### Description
This PR is a workaround where users cannot proceed with the application form if the hCaptcha client fails to load. In the current case, the hCaptcha API client isn't setting the correct CORS headers. 

Previously, the challenge request was being invoked when the form is submitted. The CORS issue above prevented the form submission from proceeding. I have moved the challenge request to a new `onLoad` handler function so that it is executed when the API client loads. If the API client does not load successfully, the challenge response value will be empty, the empty response is logged, but the application form will proceed as normal.

[AB#2977](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/2977)

### Checklist

### Test Instructions
Run the application and navigate to `/en/apply`. You should be able to start an application form.

### Additional notes
See [hCaptcha documentation on Local Development](https://docs.hcaptcha.com/#local-development) to get around the CORS issue when developing locally.

### DIFF
--- a/frontend/app/routes/$lang+/_public+/apply+/index.tsx
+++ b/frontend/app/routes/$lang+/_public+/apply+/index.tsx
@@ -70,11 +70,15 @@ export default function ApplyIndex() {
   const isSubmitting = fetcher.state !== 'idle';
   const captchaRef = useRef<HCaptcha>(null);
 
-  async function handleSubmit(event: FormEvent<HTMLFormElement>) {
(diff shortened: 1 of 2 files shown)


Fbasham (approved):
I'm going to approve this since it works (for now), plus without it we're stuck in the mud.

PR #804, opened 2024-03-26
featureEnabled function updated AB#3025

CanadianPhilipBarkow:
### Description
To reduce code repetition the function `featureEnabled` has been updated to return 404 pages if they are associated with features that are not flagged for inclusion.

[AB#3025](https://dev.azure.com/DTS-STN/Canada%20Dental%20Care%20Plan/_workitems/edit/3025)

### Checklist

- [X] I have checked that my code follows the project's coding style by running `npm run format:check`
- [X] I have checked that my code contains no linting errors by running `npm run lint`
- [X] I have checked that my code contains no type errors by running `npm run typecheck`
- [X] I have checked that all unit tests pass by running `npm run test:unit -- run`
- [X] I have checked that all e2e tests pass by running `npm run test:e2e`

### Test Instructions

1. Attempt to access a flagged feature (e.g. view letters)
2. Confirm that it can be accessed
3. Update the .env to remove the flag for that feature
4. Confirm that the feature can no longer be accessed

### Additional Notes
The .env.example file has been updated.

### DIFF
--- a/frontend/app/utils/env.server.ts
+++ b/frontend/app/utils/env.server.ts
@@ -5,6 +5,8 @@ import { z } from 'zod';
 import { privateKeyPemToCryptoKey, publicKeyPemToCryptoKey } from './crypto-utils.server';
 import { getLogger } from '~/utils/logging.server';
 
+const log = getLogger('env.server');
+
 // none, error, warn, info, debug, verbose, all
 const otelLogLevels = Object.keys(DiagLogLevel).map((key) => key.toLowerCase());
 
@@ -179,7 +181,6 @@ export function getPublicEnv() {
 
   return result.data;
 }
-const log = getLogger('env.server');
 
 /**
  * A helper function to be used in actions and loaders that checks if a feature
--- a/frontend/app/utils/env.server.ts
+++ b/frontend/app/utils/env.server.ts
@@ -184,7 +184,9 @@ export function getPublicEnv() {
  */
 export function featureEnabled(feature: FeatureName) {
   const { ENABLED_FEATURES } = getEnv();
-  return ENABLED_FEATURES.includes(feature);
+  if (!ENABLED_FEATURES.includes(feature)) {
+    throw new Response('Not Found', { status: 404 });
+  }
 }
 
 /**
--- a/frontend/app/utils/env.server.ts
+++ b/frontend/app/utils/env.server.ts
@@ -3,6 +3,7 @@ import { randomUUID } from 'node:crypto';
 import { z } from 'zod';
 
 import { privateKeyPemToCryptoKey, publicKeyPemToCryptoKey } from './crypto-utils.server';
+import { getLogger } from '~/utils/logging.server';
 
 // none, error, warn, info, debug, verbose, all
 const otelLogLevels = Object.keys(DiagLogLevel).map((key) => key.toLowerCase());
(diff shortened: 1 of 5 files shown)


sebastien-comeau:
"higher order functions" is not recommended by Remix.

https://remix.run/docs/en/main/guides/constraints#higher-order-functions

gregory-j-baker:
> "higher order functions" is not recommended by Remix.
> 
> https://remix.run/docs/en/main/guides/constraints#higher-order-functions

It was me who suggested using a HOF. The pruning issue mentioned in the link above is valid, and something I never considered.

That pruning issue can be mitigated by ensuring that the route loader doesn't produce a side effect, but the code required to do that doesn't differ that much from the code that would be required to just explicitly call your function. So in this case, I think the best thing to do would be to:

1. move your function into a `.server.ts` file
2. have it throw a 404 response if the feature is not enabled (NOOP if the feature is enabled)
3. Unwrap all the loaders and put the call to your function as the first line in each.

Sorry about this. I didn't realize that HOFs were potentially problematic in Remix.

frontend/app/utils/env.server.ts
```diff
@@ -181,7 +181,9 @@ export function getPublicEnv() {
  */
 export function featureEnabled(feature: FeatureName) {
   const { ENABLED_FEATURES } = getEnv();
-  return ENABLED_FEATURES.includes(feature);
+  if (!ENABLED_FEATURES.includes(feature)) {
+    throw new Response('Not Found', { status: 404 });
+  }
 }
```
gregory-j-baker, line 187: I would modify this implementation slightly (update jsdoc; log a warning)..

```typescript
const log = getLogger('env.server');

/**
 * A helper function to be used in actions and loaders that checks if a feature
 * is enabled. If the feature is not enabled, a 404 response will be thrown.
 */
export function featureEnabled(feature: FeatureName) {
  const { ENABLED_FEATURES } = getEnv();

  if (!ENABLED_FEATURES.includes(feature)) {
    log.warn('Feature [%s] is not enabled; returning 404 response', feature);
    throw new Response(null, { status: 404 });
  }
}
```

gregory-j-baker (dismissed):
Approve, but with a small suggested change.

frontend/app/utils/env.server.ts
gregory-j-baker: **Reason:** it might be helpful to future devs when they hit a URL and get a 404.

gregory-j-baker: Also: I removed the response body.. Browsers understand a 404, so there's no need to send a 'Not Found' response body.

CanadianPhilipBarkow: Makes sense. Implemented in commit 6278670f1110c2b2127cd93ac3535624f55c6078.

frontend/app/utils/env.server.ts
```diff
@@ -178,13 +179,18 @@ export function getPublicEnv() {
 
   return result.data;
 }
+const log = getLogger('env.server');
```
gregory-j-baker, line 182: We usually put this at the top of the file (after the imports) so it can be used by any/all functions.

gregory-j-baker (changes requested):

frontend/app/utils/env.server.ts
CanadianPhilipBarkow: Also makes sense. Implemented in commit 6de167b3b81f6af6f5d21ba1c58a95794b273d32.

gregory-j-baker (approved):

PR #818, opened 2024-03-26
if letter.name is invalid, do not display the letter.

nibivi77:
### Description
If letter.name is invalid, do not display the letter.

This check can also be done in the `/app/services` `getLetters()`

[AB#3144](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/3144)

### Checklist

### Test Instructions
Navigate to /en/letters
Nothing should change since there are currently no letters in our mocks that have an invalid name.

### Additional Notes
Once we get the new JSON file, some tweaks might be needed. However, the structure of the new JSON file is said to not alter the current one.

### DIFF
--- a/frontend/app/routes/$lang+/_protected+/letters+/index.tsx
+++ b/frontend/app/routes/$lang+/_protected+/letters+/index.tsx
@@ -97,13 +97,14 @@ export default function LettersIndex() {
       <ul className="divide-y border-y">
         {letters.map((letter) => {
           const letterType = letterTypes.find(({ id }) => id === letter.name);
-          const translatedLetterName = letterType ? getNameByLanguage(i18n.language, letterType) : letter.name;
-          const letterName = translatedLetterName ?? letter.name;
+          if (!letterType) {
+            return;
+          }
 
           return (
             <li key={letter.id} className="py-4 sm:py-6">
               <InlineLink reloadDocument to={`/letters/${letter.id}/download`} className="external-link font-lato font-semibold" target="_blank">
-                {letterName}
+                {getNameByLanguage(i18n.language, letterType)}
               </InlineLink>
               <p className="mt-1 text-sm text-gray-500">{t('letters:index.date', { date: letter.issuedOn })}</p>
             </li>


frontend/app/routes/$lang+/_protected+/letters+/index.tsx
```diff
@@ -100,6 +100,7 @@ export default function LettersIndex() {
       </div>
       <ul className="divide-y border-y">
         {letters.map((letter) => {
+          if (letter.name === undefined || letter.name === '') return;
```
gregory-j-baker, line 103: A cleaner way to do this would be to use `.filter(..)`:

``` tsx
letters
  .filter(letter => letter.name?.length > 0)
  .map(..)
```

But if all of your letters have undefined or empty names, you will be left with an empty `<ul></ul>`... so my suggestion would be to filter inside of the component code (ie: before the `return`) and then conditionally render either the `<ul>` or `<p>There are no letters on file.</p>` (or something similar).

gregory-j-baker (dismissed):

frontend/app/routes/$lang+/_protected+/letters+/index.tsx
```diff
@@ -100,6 +100,7 @@ export default function LettersIndex() {
...
           const letterType = letterTypes.find(({ id }) => id === letter.name);
           const translatedLetterName = letterType ? getNameByLanguage(i18n.language, letterType) : letter.name;
           const letterName = translatedLetterName ?? letter.name;
```
nicholas-ly, line 106: With this newly introduced code, if a `letter.name` is found but does not exist in `letterTypes`, we are going to default to displaying the `letter.name` (ie. the ID of the letter type). 

However, the user story indicates if the `letter.name` is not in the Power Platform mock (ie. the value of `letterTypes`), do not display the letter.

I think we should be adding an `if (!letterType) { return <></> }` check instead of what is introduced in this PR.

gregory-j-baker: One could make a case for handling some of this mapping inside of the service module. Why not just return a collection of letters with the letter type already pre-mapped. If no type exists, map it to `undefined` and then filter that out during rendering with a `.filter(letter.name && letter.type)` sort of thing.

nibivi77: Let me know if you prefer to map this in the service instead.

nicholas-ly (approved):

PR #839, opened 2024-03-28
Creating sceean reader new tab component.

nibivi77:
### Description
Reusable span sr-only component.

[AB#3164](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/3164)

### Checklist

### Test Instructions
Navigate to `/letters` and inspect each link in the list. <span sr-only> should be right after the text.

### DIFF
--- a/frontend/app/components/new-tab-indicator.tsx
+++ b/frontend/app/components/new-tab-indicator.tsx
@@ -1,0 +1,14 @@
+import type { ComponentProps } from 'react';
+
+import { useTranslation } from 'react-i18next';
+
+import { cn } from '~/utils/tw-utils';
+
+export function NewTabIndicator({ className, ...props }: Omit<ComponentProps<'span'>, 'children'>) {
+  const { t } = useTranslation('gcweb');
+  return (
+    <span className={cn('sr-only', className)} {...props}>
+      {/*Following whitespace is important to ensure the content's text is seperated for the screen-reader text*/}&#32;({t('screen-reader.new-tab')})
+    </span>
+  );
+}
--- a/frontend/app/routes/$lang+/_protected+/letters+/index.tsx
+++ b/frontend/app/routes/$lang+/_protected+/letters+/index.tsx
@@ -11,6 +11,7 @@ import pageIds from '../../page-ids.json';
 import { ButtonLink } from '~/components/buttons';
 import { InlineLink } from '~/components/inline-link';
 import { InputSelect } from '~/components/input-select';
+import { NewTabIndicator } from '~/components/new-tab-indicator';
 import { getAuditService } from '~/services/audit-service.server';
 import { getInstrumentationService } from '~/services/instrumentation-service.server';
 import { getLettersService } from '~/services/letters-service.server';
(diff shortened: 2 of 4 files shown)


frontend/app/routes/$lang+/_protected+/letters+/index.tsx
```diff
@@ -105,6 +106,7 @@ export default function LettersIndex() {
             <li key={letter.id} className="py-4 sm:py-6">
               <InlineLink reloadDocument to={`/letters/${letter.id}/download`} className="external-link font-lato font-semibold" target="_blank">
                 {getNameByLanguage(i18n.language, letterType)}
+                <NewTabIndicator>{t('gcweb:screen-reader.newtab')}</NewTabIndicator>
```
sebastien-comeau, line 109: Move the text translation inside the `NewTabIndicator` component, otherwise that component is basically useless.

frontend/public/locales/en/gcweb.json
```diff
@@ -122,5 +122,8 @@
       "label": "Year",
       "placeholder": "YYYY"
     }
+  },
+  "screen-reader": {
+    "newtab": "Opens in a new tab"
```
sebastien-comeau, line 127: Should be in lowercase.

```suggestion
    "newtab": "opens in a new tab"
```

frontend/public/locales/fr/gcweb.json
```diff
@@ -122,5 +122,8 @@
       "label": "Année",
       "placeholder": "AAAA"
     }
+  },
+  "screen-reader": {
+    "newtab": "S'ouvre dans un nouvel onglet"
```
sebastien-comeau, line 127: Should be in lowercase

```suggestion
    "newtab": "s'ouvre dans un nouvel onglet"
```

frontend/app/components/new-tab-indicator.tsx
```diff
@@ -0,0 +1,10 @@
+import type { ComponentProps, ReactNode } from 'react';
+
+export interface NewTabIndicatorProps extends ComponentProps<'span'> {
+  children: ReactNode;
+}
```
sebastien-comeau, line 5: Omit children prop since we will handle the content in the component.

```suggestion
export interface NewTabIndicatorProps extends Omit<ComponentProps<'span'>, 'children'> { }
```

frontend/app/components/new-tab-indicator.tsx
```diff
@@ -0,0 +1,10 @@
...
+
+export function NewTabIndicator(props: NewTabIndicatorProps) {
+  const { children } = props;
+  return <span className="sr-only">{children}</span>;
+}
```
sebastien-comeau, line 10: Handle the content here. `&#32` is important to ensure the content's text is seperated for the `opens in a new tab` text.

```suggestion
export function NewTabIndicator({className, ...props}: NewTabIndicatorProps) {
  const { t } = useTranslation("gcweb");
  return <span className={cn("sr-only", className)} ...props>&#32;({t('gcweb:screen-reader.newtab')})</span>;
}
```

sebastien-comeau (changes requested):

frontend/app/components/new-tab-indicator.tsx
gregory-j-baker: Just use a regular space, and consider adding a comment for future devs so it isn't removed..

```
// note: leading space is required for screen readers to correctly identify word boundaries
<span ..> ({t('..')})</span>
```

something like that

sebastien-comeau: @gregory-j-baker I think to be safe we should explicitly set it `&#32`. I don't want Prettier to format the file and remove it.

sebastien-comeau (approved):

frontend/app/components/new-tab-indicator.tsx
gregory-j-baker: @sebastien-comeau always has to disagree 🏃🏻

PR #960, opened 2024-04-09
3257 fix status checker link on application submitted page

Fbasham:
### Description
status checker link wasn't navigating correctly because it was missing the locale

[AB#3257](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/3257)

### DIFF
--- a/frontend/app/routes/$lang+/_public+/apply+/$id+/confirmation.tsx
+++ b/frontend/app/routes/$lang+/_public+/apply+/$id+/confirmation.tsx
@@ -1,6 +1,6 @@
 import type { ActionFunctionArgs, LoaderFunctionArgs, MetaFunction } from '@remix-run/node';
 import { json, redirect } from '@remix-run/node';
-import { useFetcher, useLoaderData } from '@remix-run/react';
+import { useFetcher, useLoaderData, useParams } from '@remix-run/react';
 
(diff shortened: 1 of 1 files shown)


frontend/app/routes/$lang+/_public+/apply+/$id+/confirmation.tsx
```diff
@@ -155,7 +155,7 @@ export default function ApplyFlowConfirm() {
 
   const mscaLink = <InlineLink to={t('confirm.msca-link')} />;
   const dentalContactUsLink = <InlineLink to={t('confirm.dental-link')} />;
-  const cdcpLink = <InlineLink to={'/status'} />;
+  const cdcpLink = <InlineLink to={`/${i18n.language}/status`} />;
```
gregory-j-baker, line 158: This is not the correct way to link to internal pages anymore. Do this instead:

``` tsx
<InlineLink routeId="$lang+/_public+/status+/index" params={params} />
```

Where `params` is initialized via `useParams()`.

This is because all internal paths are now bilingual, and we need to link to the correct one depending on the current language.

gregory-j-baker (changes requested):

gregory-j-baker (approved):

PR #1019, opened 2024-04-11
Only transform pageLoad url for specific pathnames

sebastien-comeau:
### Description

Only transform pageLoad url for specific pathnames.

[AB#2830](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/2830)

### Checklist

### DIFF
--- a/frontend/app/utils/adobe-analytics.client.ts
+++ b/frontend/app/utils/adobe-analytics.client.ts
@@ -1,5 +1,3 @@
-import { removeUUIDSegmentsFromURL } from './url-utils';
-
 type AdobeDataLayer = { push?: (object: Record<string, string | Record<string, string>>) => void };
 
 declare global {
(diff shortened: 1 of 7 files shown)


frontend/app/utils/adobe-analytics.client.ts
```diff
@@ -24,16 +24,22 @@ export const pageview = (locationUrl: string) => {
...
   // website visitor categories.
-  const transformedUrl = removeUUIDSegmentsFromURL(locationUrlObj);
+  if (
+    ['/en/apply/', '/fr/apply/', '/en/appliquer/', '/fr/appliquer/'].some((str) => {
+      return locationUrlObj.pathname.toLocaleLowerCase().startsWith(str);
+    })
+  ) {
+    const transformedUrl = removeUUIDSegmentsFromURL(locationUrlObj);
+    url = `${transformedUrl.host}${transformedUrl.pathname}`;
+  }
```
gregory-j-baker, line 39: I don't like that the URL removal code is so specific.. like mentioned in slack, if the user changes the UUID in the address to something that isn't valid (ex: removes a hyphen), then the `removeUUIDSegmentsFromURL()` call will not work.

Also, since this code is specific to the apply flow, it should be colocated with it. This code isn't globally usable, so there's no reason to put it in a util module.

My suggestion would be to move the code into the /appy/ route, and have it remove the flow ID by doing something like `replace(/(en|fr)\/(apply|appliquer)\/.+(\/.*)/$1$2$3)`

gregory-j-baker (changes requested):

frontend/app/utils/adobe-analytics.client.ts
sebastien-comeau: I completely revamped that code. Please review again.

gregory-j-baker (approved):

PR #1154, opened 2024-04-23
[gitops] Update `prod` to include required manifests

gregory-j-baker:
In preparation for the May 01 go-live, I have updated the production configuration files with the required manifests.

**Note 01:** an OAuth Proxy authwall has been added to keep the site inaccessible until May 01. I have annotated the yaml files with comments anywhere OAuth Proxy is used so that it will be easier to remove it when we go live.

**Note 02:** I have deployed a pre-release build (v0.0.0) to test everything. The deployed version will have to be updated once we go live.


gitops/overlays/prod/configs/frontend/config.conf
```diff
@@ -0,0 +1,51 @@
+SCCH_BASE_URI=https://service.canada.ca
+MSCA_BASE_URI=https://srv136.services.gc.ca
```
nicholas-ly, line 2: These values are the default values that are set in the application - no need to explicitly set them in this file.

gitops/overlays/prod/configs/frontend/config.conf
```diff
@@ -0,0 +1,42 @@
...
+SESSION_STORAGE_TYPE=redis
+REDIS_SENTINEL_NAME=myprimary
+REDIS_SENTINEL_HOST=redis
+REDIS_SENTINEL_PORT=26379
+
+#
+# RAOIDC configuration
+#
+AUTH_LOGOUT_REDIRECT_URL=https://srv136.services.gc.ca/ecas-seca/raoidc/v1/logout?client_id={clientId}&shared_session_id={sharedSessionId}&ui_locales={uiLocales}
+AUTH_RAOIDC_BASE_URL=https://srv136.services.gc.ca/ecas-seca/raoidc/v1/
```
nicholas-ly, line 25:
```suggestion
AUTH_RAOIDC_BASE_URL=https://srv136.services.gc.ca/ecas-seca/raoidc/v1
```

gitops/overlays/prod/configs/frontend/config.conf
gregory-j-baker: Fixed that.

gitops/overlays/prod/configs/frontend/config.conf
gregory-j-baker: Fixed and force-pushed.

PR #1170, opened 2024-04-24
prevent user from navigating if they aren't of proper age

Fbasham:
### Description
Prevent users from navigating to certain screens in apply flow if their current age isn't what it should be to access the screen.

### DIFF
--- a/frontend/app/routes/$lang+/_public+/apply+/$id+/living-independently.tsx
+++ b/frontend/app/routes/$lang+/_public+/apply+/$id+/living-independently.tsx
@@ -15,6 +15,7 @@ import { InputRadios } from '~/components/input-radios';
 import { Progress } from '~/components/progress';
 import { getApplyRouteHelpers } from '~/route-helpers/apply-route-helpers.server';
 import * as adobeAnalytics from '~/utils/adobe-analytics.client';
+import { getUserAge } from '~/utils/date-utils';
 import { getTypedI18nNamespaces } from '~/utils/locale-utils';
 import { getFixedT } from '~/utils/locale-utils.server';
 import { getLogger } from '~/utils/logging.server';
@@ -47,6 +48,11 @@ export async function loader({ context: { session }, params, request }: LoaderFu
 
   const csrfToken = String(session.get('csrfToken'));
   const meta = { title: t('gcweb:meta.title.template', { title: t('apply:living-independently.page-title') }) };
+
+  const age = getUserAge(state.dateOfBirth ?? '');
+  if (age !== 16 && age !== 17) {
+    return redirect(getPathById('$lang+/_public+/apply+/$id+/date-of-birth', params));
+  }
 
   return json({ id: state.id, csrfToken, meta, defaultState: state.livingIndependently });
 }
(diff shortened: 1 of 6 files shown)


frontend/app/routes/$lang+/_public+/apply+/$id+/living-independently.tsx
```diff
@@ -48,6 +49,12 @@ export async function loader({ context: { session }, params, request }: LoaderFu
   const csrfToken = String(session.get('csrfToken'));
   const meta = { title: t('gcweb:meta.title.template', { title: t('apply:living-independently.page-title') }) };
 
+  const parseDateOfBirth = parse(state.dateOfBirth ?? '', 'yyyy-MM-dd', new Date());
+  const age = differenceInYears(new Date(), parseDateOfBirth);
```
gregory-j-baker, line 53: Two things..

1. what does `new Date()` do in the `parse()` call?
2. the pods have their timezones set to UTC.. which means there could potentially be an off-by-one error when using `new Date()`.

When you need to get today's *date* (as opposed to the current *instant*), you should probably specify the timezone.

Newfoundland time is a pretty good timezone, I've heard.

gregory-j-baker (commented):
@nicholas-ly anything to add about the dates?

frontend/app/routes/$lang+/_public+/apply+/$id+/living-independently.tsx
Fbasham: 1.  it comes from the `date-fns` package: ```* `referenceDate` must be passed for correct work of the function.
 * If you're not sure which `referenceDate` to supply, create a new instance of Date:
 * `parse('02/11/2014', 'MM/dd/yyyy', new Date())`
 * In this case parsing will be done in the context of the current date.
 * If `referenceDate` is `Invalid Date` or a value not convertible to valid `Date`,
 * then `Invalid Date` will be returned.```

I pulled the logic from `date-of-birth.tsx`.  Since this same conditional is being used n a few places after submitting the date of birth screen in the apply flow, what I'm inclined to do is calculate the age at the time of submission of `date-of-birth.tsx` in state and just use value to do the conditional checks.

nicholas-ly: > @nicholas-ly anything to add about the dates?

There can be a off-by-one discrepancy in age calculations due to time zone differences even if we use a consistent time zone like UTC. 

@gregory-j-baker suggested using the Newfoundland time zone (easternmost in Canada) as the base to limit the discrepancy across Canada. This wouldn't entirely eliminate the issue though but would reduce the amount of time in which this can happen.

Fbasham: The exact same code is already in the application (https://github.com/DTS-STN/canadian-dental-care-plan/blob/0a878fdb8311de5a00df98e3134ee98ca6df293d/frontend/app/routes/%24lang%2B/_public%2B/apply%2B/%24id%2B/date-of-birth.tsx#L153)

I can put this in, but I don't see why now, all of a sudden, we care about timezones.

leiye29 (approved):

frontend/app/routes/$lang+/_public+/apply+/$id+/living-independently.tsx
Fbasham: Okay..so let's make some assumptions:  the applicant will be Canadian.  We have a lower and upper bound on what timezones can be in Canada.  We can compute the users age based on those upper and lower bounds.  If the user's age in range of those bounds (inclusive), then they meet the age criteria for the conditional.  What about that approach?

PR #1204, opened 2024-04-26
Banner text change.

nibivi77:
### Description
Banner text change.


nibivi77:
Needs to be on release and not main.

PR #1303, opened 2024-05-07
Alert page. Adding text when new code requested.

nibivi77:
### Description
Adding text when new code requested.

[AB#3349](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/3349)
### Checklist

### DIFF
--- a/frontend/app/routes/$lang+/_protected+/alerts+/subscribe+/confirm.tsx
+++ b/frontend/app/routes/$lang+/_protected+/alerts+/subscribe+/confirm.tsx
@@ -55,6 +55,11 @@ export async function loader({ context: { session }, params, request }: LoaderFu
 
   const userInfoToken: UserinfoToken = session.get('userInfoToken');
   const alertSubscription = await getSubscriptionService().getSubscription(userInfoToken.sin ?? '');
+  const log = getLogger('alerts-subscription.confirm');
+  if (!alertSubscription) {
+    log.warn('Alert Subscription could not be found; responding with 404');
+    throw new Response(null, { status: 404 });
+  }
   session.set('alertSubscription', alertSubscription);
 
   const confirmationCodeEntered = session.get('codeEntered') ?? '';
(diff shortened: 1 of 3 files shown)


frontend/app/routes/$lang+/_protected+/alerts+/subscribe+/confirm.tsx
```diff
@@ -184,8 +184,7 @@ export default function ConfirmSubscription() {
                 </div>
               </>
             ) : (
-              //TODO add content: https://dev.azure.com/DTS-STN/Canada%20Dental%20Care%20Plan/_workitems/edit/3349
-              'Requested'
+              <p>{t('alerts:confirm.code-sent-by-email', { userEmailAddress: alertSubscription?.email })}</p>
```
gregory-j-baker, line 187: The optional chaining suggests that the email can be null/undefined here. Since the rendered text won't make sense if it is, I think you should wrap the entire block in a conditional and only render when `email` is defined.

gregory-j-baker (changes requested):

frontend/app/routes/$lang+/_protected+/alerts+/subscribe+/confirm.tsx
```diff
@@ -166,28 +171,31 @@ export default function ConfirmSubscription() {
...
-                <p id="confirmation-information" className="mb-4">
-                  <Trans ns={handle.i18nNamespaces} i18nKey="alerts:confirm.confirmation-information-text" values={{ userEmailAddress: alertSubscription?.email }} />
-                </p>
-                <p id="confirmation-completed" className="mb-4">
-                  {t('alerts:confirm.confirmation-completed-text')}
-                </p>
-                <div className="grid gap-12 md:grid-cols-2">
-                  <p id="confirmation-language" className="mb-4 font-bold">
-                    {t('alerts:confirm.confirmation-selected-language')}
+          {!undefinedEmail && (
```
gregory-j-baker, line 174: I don't think you need to define a new variable for this. Wouldn't this work?

``` tsx
{ alertSubscription?.email && (
  <p>🦄</p>
)}
```

gregory-j-baker (changes requested):

frontend/app/routes/$lang+/_protected+/alerts+/subscribe+/confirm.tsx
sebastien-comeau: `alertSubscription` should not be returned as `underfined` from the loader. If alertSubscription is undefined then the page should redirect or return a 404.

gregory-j-baker: I do kind of agree with @sebastien-comeau here.. I didn't think about the larger context.

Under what situation could this email become undefined, and does it make sense to render anything at all in that situation?

nibivi77: Added the block of code to throw a 404.

gregory-j-baker (approved):

PR #1316, opened 2024-05-09
[backend] Minor code fixups

gregory-j-baker:
### Minor code fixups

- Configured h2 to load when no other database is configured
- Cleaned up some formatting
- Added ApplicationIT (spring boot integration test)
- Removed SampleController


backend/src/main/java/ca/gov/dtsstn/cdcp/api/controller/SubscriptionControllerV1.java
```diff
@@ -15,7 +16,7 @@
 public class SubscriptionControllerV1 {
 
 	@GetMapping({ "/users/{userId}/subscriptions" })
-	public List<Subscription> getSubscriptions(@PathVariable String userId) {
+	public Collection<Subscription> getSubscriptions(@PathVariable String userId) {
```
sebastien-comeau, line 19: Since it returns a specific implementation of the Collection interface (List), it's more appropriate to declare the return type as List<Subscription> to reflect the specific type being returned accurately.

```suggestion
	public List<Subscription> getSubscriptions(@PathVariable String userId) {
```

sebastien-comeau (changes requested):

sebastien-comeau (approved):

PR #1348, opened 2024-05-11
AB#3571 SYSTEST - BUG in communication preference

shawn320:
### Description
When the email is pre-populated from the previous page, the validation error still gets triggered
Updated the validation logic for email field to check if the field is pre-populated

[AB#3571](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/3571)

### Checklist


sebastien-comeau:
I will review it because I don't think the validation should change. I will debug it and see if your solution is the way to go.

sebastien-comeau (commented):
I've tested locally with other possible solution and the way it should be handled is to use `readOnly` prop on the inputs instead of `disabled`. FormData doesn't returns disabled fields but read-only fields will be returned.

sebastien-comeau (approved):

PR #1433, opened 2024-05-23
3699 Bug with /adult/partner-information

Fbasham:
### Description
Returning `fatal: true` in the zod `ctz.addIssue` causes the validation to abort early.  The SIN check is done before validating the date, which is why this "bug" happens.  By removing `fatal: true`, the validation can continue on to the date check in the `superRefine`, and everything should work as expected.

[AB#3699](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/3699)

image


sebastien-comeau (approved):
Good catch.

PR #1441, opened 2024-05-23
Added Confirmation Code Verification Endpoint

CanadianPhilipBarkow:
### Description
Confirmation code verification endpoint now ended. Updates done to `UserService` and `UsersController` classes.

[AB#3371](https://dev.azure.com/DTS-STN/Canada%20Dental%20Care%20Plan/_workitems/edit/3371)

### Checklist

- [X] I have added/updated tests that prove my fix is effective or that my feature works
- [X] I have checked that my code follows the project's coding style by running `npm run format:check`
- [X] I have checked that my code contains no linting errors by running `npm run lint`
- [X] I have checked that my code contains no type errors by running `npm run typecheck`
- [X] I have checked that all unit tests pass by running `npm run test:unit -- run`
- [X] I have checked that all e2e tests pass by running `npm run test:e2e`

### Test Instructions
1. Run updated Junit tests


backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
```diff
@@ -50,4 +54,21 @@ public UserModel getUserById(
...
+
+    @PostMapping({ "/verifyCode/"})
+    @Operation(summary = "Verify the status of a confirmation code.", operationId="verify-confirmation-code")
+    public ConfirmationCodeStatus getConfirmationCodeStatus(
+            @NotBlank(message = "code must not be null or blank")
+            @Parameter(description = "The confirmation code.", required = true)
+            @RequestParam ConfirmationCode code,
+            @NotBlank(message = "userId must not be null or blank")
+            @Parameter(description = "The ID of the user.", required = true)
+            @RequestParam String userId
```
nicholas-ly, line 66: POST requests should accept a request body rather than request/query parameters, which are generally used for filtering resources.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
```diff
@@ -50,4 +54,21 @@ public UserModel getUserById(
 		return userModelAssembler.toModel(user);
 	}
 
+
+    @PostMapping({ "/verifyCode/"})
+    @Operation(summary = "Verify the status of a confirmation code.", operationId="verify-confirmation-code")
+    public ConfirmationCodeStatus getConfirmationCodeStatus(
+            @NotBlank(message = "code must not be null or blank")
+            @Parameter(description = "The confirmation code.", required = true)
+            @RequestParam ConfirmationCode code,
```
nicholas-ly, line 63: Should `code` be a `String` rather than the entire `ConfirmationCode` entity? From the POV of a possible API consumer like a front-end web application, I wouldn't know what to pass for the other fields (like `expiryDate`, `lastModifiedBy`) of the `ConfirmationCode` aside from the code itself.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/service/UserService.java
```diff
@@ -134,4 +134,16 @@ private Predicate<ConfirmationCodeEntity> byCode(String code) {
 		return confirmationCode -> confirmationCode.getCode().equals(code);
 	}
 
+    public ConfirmationCodeStatus verifyConfirmationCode(ConfirmationCode confirmationCode, User user){
+        if (confirmationCode == null || confirmationCode.getCode().isBlank()) {
+            return ConfirmationCodeStatus.NO_CODE;
+        }
```
nicholas-ly, line 140: We should throw early here (with an `Assert.hasText(..)`) rather than returning `NO_CODE` as we should always expect a confirmation code to be passed in order to verify it.

backend/src/test/java/ca/gov/dtsstn/cdcp/api/service/UserServiceTests.java
```diff
@@ -85,4 +90,49 @@ void testCreateConfirmationCodeForUser() {
 		assertThat(confirmationCode.getExpiryDate()).isAfter(Instant.now());
 	}
 
+	@Test
+	@DisplayName("Test userService.verifyConfirmationCode(..)")
+	void testVerifyConfirmationCode_verified() {
+		final Instant creationDate = LocalDateTime.now().minusDays(73).toInstant(ZoneOffset.UTC);
+		final Instant expiryDate = LocalDateTime.now().plusDays(288).toInstant(ZoneOffset.UTC);
+		final ConfirmationCode confirmationCode = ImmutableConfirmationCode.builder().code("code value").createdDate(creationDate).expiryDate(expiryDate).build();
```
nicholas-ly, line 98: Consider using `var` for these local variable declarations (same comment for the other tests)

gregory-j-baker (changes requested):
A good attempt, but not how this should be implemented. As stated in the various comments, you should:

- Accept a POST to `/api/v1/users/{userId}/email-validations` and return either a 202 or a 400 to indicate success/failure
- When the incoming confirmation code matches one we have on file, update the `emailValidated` field of the user and delete all confirmation codes for that user
- Make the service method as simple as possible.. if it only needs a code and a userId, then only accept those. Don't require the caller to build complex objects unless the number of parameters you have is large.

gregory-j-baker:
My suggestion for your future PRs is to keep them lean and clean. You don't have to submit PRs with a full blown implementation. If you do this and make some incorrect assumptions, you end up with dozens of comments and requested changes.

This PR should have been a single service class that is used to validate a confirmation code, along with a unit test to prove that it does what it should.

Then, the next PR could have been the controller class that implements the API endpoint.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/service/UserService.java
```diff
@@ -134,4 +134,25 @@ private Predicate<ConfirmationCodeEntity> byCode(String code) {
 		return confirmationCode -> confirmationCode.getCode().equals(code);
 	}
 
+	private Predicate<ConfirmationCodeEntity> notExpired() {
+		return confirmationCode -> confirmationCode.getExpiryDate().isAfter(Instant.now());
+	}
+
+    public Boolean verifyConfirmationCode(String code, User user){
```
gregory-j-baker, line 141: Pass in the confirmation code and the user id. Don't force the caller to construct a `User` object when a simple string would be sufficient.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/service/UserService.java
```diff
@@ -134,4 +134,25 @@ private Predicate<ConfirmationCodeEntity> byCode(String code) {
...
+		Assert.hasText(code, "code is required; it must not be null or blank");
+		Assert.notNull(user, "user is required; it must not be null");
+		if (user.getConfirmationCodes().stream().map(confirmationCodeMapper::toEntity).filter(notExpired()).noneMatch(byCode(code))) {
+    		return false;
+    	}
+        return true;
```
gregory-j-baker, line 147: When a confirmation code is validated, all confirmation codes for the user must be deleted from the database and the user's `emailValidated` flag must be set to `true`.

gregory-j-baker, line 147:
``` java
return user.getConfirmationCodes().stream()
  .filter(byCode(code)).findFirst()
  .isPresent();
```

backend/src/main/java/ca/gov/dtsstn/cdcp/api/service/UserService.java
```diff
@@ -134,4 +134,25 @@ private Predicate<ConfirmationCodeEntity> byCode(String code) {
...
+    }
+
+	public void setEmailValidated(String userId) {
```
gregory-j-baker, line 150: I would probably roll this method into the verify method above and name it `verifyEmail(..)` since searching for the code, setting verified=true, and deleting all other confirmation codes will always be an atomic operation.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
```diff
@@ -50,4 +54,27 @@ public UserModel getUserById(
 		return userModelAssembler.toModel(user);
 	}
 
+
+    @PostMapping({ "/{userId}/email-validations"})
+    @Operation(summary = "Verify the status of a confirmation code.", operationId="verify-confirmation-code")
```
gregory-j-baker, line 59: The API operation description doesn't really match what the endpoint is doing. I would change this to something like "validate a user's email via a confirmation code" or something like that.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
```diff
@@ -50,4 +54,27 @@ public UserModel getUserById(
...
+    public ResponseEntity<?>  verifyConfirmationCodeStatus(
+            @NotBlank(message = "userId must not be null or blank")
+            @Parameter(description = "The ID of the user.", required = true)
+            @PathVariable String userId,
+            @NotBlank(message = "code must not be null or blank")
+            @Parameter(description = "The confirmation code.", required = true)
+            @RequestParam String code
```
gregory-j-baker, line 66: This should be in the POST body.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
```diff
@@ -50,4 +54,27 @@ public UserModel getUserById(
 		return userModelAssembler.toModel(user);
 	}
 
+
+    @PostMapping({ "/{userId}/email-validations"})
+    @Operation(summary = "Verify the status of a confirmation code.", operationId="verify-confirmation-code")
+    public ResponseEntity<?>  verifyConfirmationCodeStatus(
```
gregory-j-baker, line 60: This can return `void` with a `@ResponseStatus` annotation.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
gregory-j-baker: To verify that the code is valid, use a validator. This allows the endpoint to return 204 when the code is accepted, and 400 when it is rejected.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
```diff
@@ -50,4 +54,27 @@ public UserModel getUserById(
...
+            @NotBlank(message = "userId must not be null or blank")
+            @Parameter(description = "The ID of the user.", required = true)
+            @PathVariable String userId,
+            @NotBlank(message = "code must not be null or blank")
+            @Parameter(description = "The confirmation code.", required = true)
+            @RequestParam String code
+        ){
+			final var user = userService.getUserById(userId)
+				.orElseThrow(() -> new ResourceNotFoundException("No user with id=[%s] was found".formatted(userId)));
+			Assert.hasText(code, "id is required; it must not be null or blank");
```
gregory-j-baker, line 70: 1. Assertions are usually done up front before any other code is executed.
2. There's a copy/paste error in the code.
3. The code should be sent in a request body.
4. The code should be validated using the JSR validation framework.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
```diff
@@ -50,4 +54,27 @@ public UserModel getUserById(
...
+			if (userService.verifyConfirmationCode(code, user)) {
+				userService.setEmailValidated(userId);
+				return new ResponseEntity<>(HttpStatus.ACCEPTED);
+			}
+			return new ResponseEntity<>(HttpStatus.BAD_REQUEST);
+		
+
```
gregory-j-baker, line 77:
```suggestion
```

backend/src/test/java/ca/gov/dtsstn/cdcp/api/service/UserServiceTests.java
```diff
@@ -85,4 +88,34 @@ void testCreateConfirmationCodeForUser() {
...
+	@Test
+	@DisplayName("Test userService.verifyConfirmationCode(..) with valid input")
+	void testVerifyConfirmationCode_Valid() {
+		final var creationDate = LocalDateTime.now().minusDays(73).toInstant(ZoneOffset.UTC);
+		final var expiryDate = LocalDateTime.now().plusDays(288).toInstant(ZoneOffset.UTC);
+		final var codeValue = "code value";
+		final var confirmationCode = ImmutableConfirmationCode.builder().code(codeValue).createdDate(creationDate).expiryDate(expiryDate).build();
+		final var user = ImmutableUser.builder().addConfirmationCodes(confirmationCode).build();
+
+		assertThat(userService.verifyConfirmationCode(codeValue, user)).isTrue();
```
gregory-j-baker, line 100: As mentioned in other comments.. don't pass a user object into this method. Pass a user id and make the method look that user up in the database.

backend/src/test/java/ca/gov/dtsstn/cdcp/api/service/UserServiceTests.java
```diff
@@ -85,4 +88,34 @@ void testCreateConfirmationCodeForUser() {
...
+	}
+
+	@Test
+	@DisplayName("Test userService.verifyConfirmationCode(..) with no code")
+	void testVerifyConfirmationCode_NullCodeValue() {
+		final var user = ImmutableUser.builder().build();
+		assertThrowsExactly(IllegalArgumentException.class, () -> userService.verifyConfirmationCode("", user), "code is required; it must not be null");
```
gregory-j-baker, line 107: `""` is not the same as `null` so this test isn't doing what the method name suggests it is

backend/src/test/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersControllerIT.java
```diff
@@ -61,7 +65,7 @@ void testGetUserById() throws Exception {
 		when(userService.getUserById(any())).thenReturn(Optional.of(mockUser));
 
 		mockMvc.perform(get("/api/v1/users/00000000-0000-0000-0000-000000000000"))
-			.andDo(print())
+			//.andDo(print())
```
gregory-j-baker, line 68: Why was this commented out?

backend/src/test/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersControllerIT.java
```diff
@@ -79,7 +83,7 @@ void testGetUserById_NotFound() throws Exception {
 		when(userService.getUserById(any())).thenReturn(Optional.empty());
 
 		mockMvc.perform(get("/api/v1/users/00000000-0000-0000-0000-000000000000"))
-			.andDo(print())
+			//.andDo(print())
```
gregory-j-baker, line 86: Why was this commented out?

backend/src/test/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersControllerIT.java
```diff
@@ -94,7 +98,7 @@ void testGetUserById_NotFound() throws Exception {
 	@DisplayName("Test unauthenticated GET /api/v1/users/{id}")
 	void testGetUserById_Unauthorized() throws Exception {
 		mockMvc.perform(get("/api/v1/users/00000000-0000-0000-0000-000000000000"))
-			.andDo(print())
+			//.andDo(print())
```
gregory-j-baker, line 101: Why was this commented out?

backend/src/test/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersControllerIT.java
```diff
@@ -103,8 +107,57 @@ void testGetUserById_Unauthorized() throws Exception {
 	@DisplayName("Test insufficient privilege GET /api/v1/users/{id}")
 	void testGetUserById_Forbidden() throws Exception {
 		mockMvc.perform(get("/api/v1/users/00000000-0000-0000-0000-000000000000"))
-			.andDo(print())
+			//.andDo(print())
```
gregory-j-baker, line 110: Why was this commented out?

gregory-j-baker (commented):
Please carefully review your commits before requesting others to review them. This is easy to do using the various diff tools in your IDE.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
gregory-j-baker: This is still unresolved.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/service/UserService.java
gregory-j-baker: I made a comment about using JSR validation for this, which is the standard approach when the incoming data is in the POST body. I will mark this comment as resolved and we can track the input validation via the other comment.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
```diff
@@ -50,4 +54,21 @@ public UserModel getUserById(
 		return userModelAssembler.toModel(user);
 	}
 
+
+    @PostMapping({ "/verifyCode/"})
```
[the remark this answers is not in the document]
gregory-j-baker: A new controller was not implemented so I'll leave this comment as unresolved and let @nicholas-ly decide what to do with it. I have no strong opinion either way.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/service/UserService.java
```diff
@@ -134,4 +134,16 @@ private Predicate<ConfirmationCodeEntity> byCode(String code) {
 		return confirmationCode -> confirmationCode.getCode().equals(code);
 	}
 
+    public ConfirmationCodeStatus verifyConfirmationCode(ConfirmationCode confirmationCode, User user){
```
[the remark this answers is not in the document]
gregory-j-baker: I'm not going to resolve this comment since the new code forces the caller to send in a `User` object when (in my opinion) a `String userId` would be preferred.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
```diff
@@ -50,4 +54,21 @@ public UserModel getUserById(
...
+    @Operation(summary = "Verify the status of a confirmation code.", operationId="verify-confirmation-code")
+    public ConfirmationCodeStatus getConfirmationCodeStatus(
```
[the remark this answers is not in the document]
gregory-j-baker: Not resolving; see comment about JSR validations.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
gregory-j-baker: After discussing the validator comment with @CanadianPhilipBarkow, we decided that doing the check in a JSR validator isn't ideal.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/service/UserService.java
CanadianPhilipBarkow: Makes sense, implemented in commit 2cebd2d6f0685660fa512c66c4a5da16869070e1.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/service/UserService.java
CanadianPhilipBarkow: Implemented in commit 2cebd2d6f0685660fa512c66c4a5da16869070e1, though I used `anymatch()` rather than `isPresent()`.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/service/UserService.java
CanadianPhilipBarkow: Sounds reasonable, implemented in commit 2cebd2d6f0685660fa512c66c4a5da16869070e1.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
CanadianPhilipBarkow: Sounds good, implemented in commit 2cebd2d6f0685660fa512c66c4a5da16869070e1.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
CanadianPhilipBarkow: Implemented in commit 2cebd2d6f0685660fa512c66c4a5da16869070e1.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
CanadianPhilipBarkow: Method rewritten in commit 2cebd2d6f0685660fa512c66c4a5da16869070e1..

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
CanadianPhilipBarkow: Superfluous whitespace removed in commit 2cebd2d6f0685660fa512c66c4a5da16869070e1..

backend/src/test/java/ca/gov/dtsstn/cdcp/api/service/UserServiceTests.java
CanadianPhilipBarkow: Implemented in commit 2cebd2d6f0685660fa512c66c4a5da16869070e1.

backend/src/test/java/ca/gov/dtsstn/cdcp/api/service/UserServiceTests.java
CanadianPhilipBarkow: Makes sense, tests have been expanded in commit 2cebd2d6f0685660fa512c66c4a5da16869070e1.

backend/src/test/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersControllerIT.java
CanadianPhilipBarkow: This commit was done while testing was still in progress, it seemed more convenient to reduce the output while I was investigating a failing test. They have been restored in commit 2cebd2d6f0685660fa512c66c4a5da16869070e1.

backend/src/test/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersControllerIT.java
CanadianPhilipBarkow: This commit was done while testing was still in progress, it seemed more convenient to reduce the output while I was investigating a failing test. They have been restored in commit 2cebd2d6f0685660fa512c66c4a5da16869070e1.

backend/src/test/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersControllerIT.java
CanadianPhilipBarkow: This commit was done while testing was still in progress, it seemed more convenient to reduce the output while I was investigating a failing test. They have been restored in commit 2cebd2d6f0685660fa512c66c4a5da16869070e1.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
CanadianPhilipBarkow: Implemented in commit 2cebd2d6f0685660fa512c66c4a5da16869070e1.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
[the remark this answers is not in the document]
CanadianPhilipBarkow: Implemented in commit 2cebd2d6f0685660fa512c66c4a5da16869070e1.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/service/UserService.java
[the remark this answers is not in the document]
CanadianPhilipBarkow: Implemented in commit 2cebd2d6f0685660fa512c66c4a5da16869070e1.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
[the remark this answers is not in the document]
CanadianPhilipBarkow: JSR validations not used, as per our discussion.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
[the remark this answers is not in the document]
nicholas-ly: I still see `/users/{userId}/email-validations` in `UsersController`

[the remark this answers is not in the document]
CanadianPhilipBarkow: Ah, I misread that one.

Should there also be an associated `EmailValidationsService` class?

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
nicholas-ly: Although `@RequestParam` has been changed to `@RequestBody`, it does not match the proposed format of
``` json
{
  "confirmationCode": "aabbccdd"
}
```

Consider creating a new model class for email verifications that is annotated with `@RequestBody` instead.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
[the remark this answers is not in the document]
nicholas-ly: Breaking up the `UserService` wouldn't be a bad idea for a future PR. A breakdown like this may be easier to maintain and improve cohesion:

UserService:
- createUser
- updateUser

SubscriptionService:
- createSubscriptionForUser
- updateSubscriptionForUser
- deleteSubscriptionForUser

ConfirmationCodeService:
- createConfirmationCodeForUser
- verifyEmail

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
nicholas-ly: We should return a 404 if no user with the specified `userId` is found. The current implementation would return a 500.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
nicholas-ly: Going to hijack this comment but there is also a mix of spaces and tabs for indentation. Turning on whitespace characters will help identifiy this:

image

backend/src/test/java/ca/gov/dtsstn/cdcp/api/service/UserServiceTests.java
```diff
@@ -140,4 +147,42 @@ void testGetUserById() {
...
+
+	@Test
+	@DisplayName("Test userService.verifyEmail(..) with empty code string")
+	void testVerifyConfirmationCode_EmptyCodeValue() {
+		assertThrowsExactly(IllegalArgumentException.class, () -> userService.verifyEmail("", userId), "code is required; it must not be null");
+	}
+
+	@Test
+	@DisplayName("Test userService.verifyEmail(..) with mismatched input")
+	void testVerifyConfirmationCode_Mismached() {
```
nicholas-ly, line 178: For consistency, these unit test methods should be renamed to `testVerifyEmail...()` since the method being tested has been renamed.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
gregory-j-baker: Correct. Request and response bodies should be valid JSON.
<https://www.freecodecamp.org/news/rest-api-best-practices-rest-endpoint-design-examples#1usejsonastheformatforsendingandreceivingdata>

backend/src/test/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersControllerIT.java
```diff
@@ -106,4 +108,53 @@ void testGetUserById_Forbidden() throws Exception {
...
+				.build())
+			.addConfirmationCodes(ImmutableConfirmationCode.builder().code("code value").build())
+			.build();
+
+		when(userService.getUserById(any())).thenReturn(Optional.of(mockUser));
+		when(userService.verifyEmail(anyString(), anyString())).thenReturn(true);
+
+		mockMvc.perform(post("/api/v1/users/00000000-0000-0000-0000-000000000000/email-validations")
+				.with(csrf()).header("origin", "http://localhost")
+				.content("{\"code\": \"code value\"}"))
```
gregory-j-baker, line 131: This doesn't match what the actual endpoint expects, but because the validation is broken in the controller, this test is actually a false positive.

backend/src/test/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersControllerIT.java
```diff
@@ -106,4 +108,53 @@ void testGetUserById_Forbidden() throws Exception {
...
+			.andDo(print())
+			.andExpect(status().isAccepted());
+	}
+
+	@Test
+	@DisplayName("Test authenticated POST /api/v1/users/{userid}/email-validations with invalid code")
+	@WithMockUser(roles = { "Users.Administer" })
+	void testVerifyConfirmationCodeStatus_Invalid() throws Exception {
```
gregory-j-baker, line 139: Same comments for this as the other test in this class.

gregory-j-baker (changes requested):

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
CanadianPhilipBarkow: As of commit 70d88ff21299e8b6f77785dd0b22d60b10f1a733 the `EmailVerificationModel` has been created and used in this context.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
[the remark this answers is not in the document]
CanadianPhilipBarkow: As of commit 70d88ff21299e8b6f77785dd0b22d60b10f1a733 the `EmailValidationsController` has been created. Other proposed changes have been saved for future PR(s).

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
CanadianPhilipBarkow: As of commit 70d88ff21299e8b6f77785dd0b22d60b10f1a733, a `ResourceNotFoundException`  is returned (with accompanying 404 `HttpStatus`) if the user is not found.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
CanadianPhilipBarkow: Whitespace has been cleaned up in commit 70d88ff21299e8b6f77785dd0b22d60b10f1a733. I think I got it all.

backend/src/test/java/ca/gov/dtsstn/cdcp/api/service/UserServiceTests.java
CanadianPhilipBarkow: Makes sense. Implemented in commit 70d88ff21299e8b6f77785dd0b22d60b10f1a733 .

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
```diff
@@ -65,4 +69,20 @@ public void updateUserById(
 		userService.updateUser(user.getId(), userUpdateModel.getEmail());
 	}	
 
+
+    @PostMapping({ "/{userId}/email-validations"})
+    @Operation(summary = "Validate a user's email via a confirmation code.", operationId="verify-confirmation-code")
```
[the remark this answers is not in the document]
CanadianPhilipBarkow: Sure, implemented in commit 70d88ff21299e8b6f77785dd0b22d60b10f1a733 .

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
```diff
@@ -65,4 +69,20 @@ public void updateUserById(
...
+	@ResponseStatus(code = HttpStatus.ACCEPTED)
+    public void verifyConfirmationCodeStatus(
+            @NotBlank(message = "userId must not be null or blank")
+            @Parameter(description = "The ID of the user.", required = true)
+            @PathVariable String userId,
+            @NotBlank(message = "code must not be null or blank")
+            @Parameter(description = "The confirmation code.", required = true)
+            @RequestBody String code
+        ){
```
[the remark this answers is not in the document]
CanadianPhilipBarkow: Implemented in commit 70d88ff21299e8b6f77785dd0b22d60b10f1a733.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
```diff
@@ -65,4 +69,20 @@ public void updateUserById(
...
+			if (userService.verifyEmail(code, userId)==false) {
+				throw new ResponseStatusException(HttpStatus.BAD_REQUEST, "Invalid confirmation code");
+			}
```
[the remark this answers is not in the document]
CanadianPhilipBarkow: Indentation updated in commit 70d88ff21299e8b6f77785dd0b22d60b10f1a733.

backend/src/test/java/ca/gov/dtsstn/cdcp/api/service/UserServiceTests.java
```diff
@@ -49,9 +53,12 @@ class UserServiceTests {
 
 	UserService userService;
 
+	private String userId;
+
 	@BeforeEach
 	void setUp() {
 		this.userService = new UserService(applicationProperties, alertTypeRepository, languageRepository, userRepository);
+		userId = "00000000-0000-0000-0000-000000000000";
```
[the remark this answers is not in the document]
CanadianPhilipBarkow: Sure. Update made in commit 70d88ff21299e8b6f77785dd0b22d60b10f1a733.

backend/src/test/java/ca/gov/dtsstn/cdcp/api/service/UserServiceTests.java
```diff
@@ -140,4 +147,42 @@ void testGetUserById() {
 		assertThat(userService.getUserById("00000000-0000-0000-0000-000000000000")).isNotEmpty();
 	}
 
-}
+	@Test
+	@DisplayName("Test userService.verifyEmail(..) with valid input")
+	void testVerifyConfirmationCode_Valid() {
+		final var creationDate = LocalDateTime.now().minusDays(73).toInstant(ZoneOffset.UTC);
+		final var expiryDate = LocalDateTime.now().plusDays(288).toInstant(ZoneOffset.UTC);
```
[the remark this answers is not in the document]
CanadianPhilipBarkow: Okay, as of commit 70d88ff21299e8b6f77785dd0b22d60b10f1a733 the timestamp values are now directly instantiated as `Instant` objects.

backend/src/test/java/ca/gov/dtsstn/cdcp/api/service/UserServiceTests.java
```diff
@@ -140,4 +147,42 @@ void testGetUserById() {
...
+		final var codeValue = "code value";
+		final var confirmationCode = new ConfirmationCodeEntityBuilder().code(codeValue).createdDate(creationDate).expiryDate(expiryDate).build();
+		final var mockUser = new UserEntityBuilder().confirmationCodes(Collections.singleton(confirmationCode)).build();
+
+		when(userRepository.findById(any())).thenReturn(Optional.of(mockUser));
+
+		assertThat(userService.verifyEmail(codeValue, userId)).isTrue();
```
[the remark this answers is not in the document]
CanadianPhilipBarkow: I tend to think that inline object instantiation makes it harder to follow what's happening, but I don't have a strong opinion on the matter,  so as of commit 70d88ff21299e8b6f77785dd0b22d60b10f1a733 the timestamp values are now instantiated inline.

backend/src/test/java/ca/gov/dtsstn/cdcp/api/service/UserServiceTests.java
```diff
@@ -140,4 +147,42 @@ void testGetUserById() {
...
+	}
+
+	@Test
+	@DisplayName("Test userService.verifyEmail(..) with null code")
+	void testVerifyConfirmationCode_NullCodeValue() {
+		assertThrowsExactly(IllegalArgumentException.class, () -> userService.verifyEmail(null, userId), "code is required; it must not be null");
```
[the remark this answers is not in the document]
CanadianPhilipBarkow: Implemented in commit 70d88ff21299e8b6f77785dd0b22d60b10f1a733.

Though does that mean the other three tests using the `assertThrowsExactly(...)` method are also flawed?

backend/src/test/java/ca/gov/dtsstn/cdcp/api/service/UserServiceTests.java
```diff
@@ -140,4 +147,42 @@ void testGetUserById() {
...
+	}
+
+	@Test
+	@DisplayName("Test userService.verifyEmail(..) with empty code string")
+	void testVerifyConfirmationCode_EmptyCodeValue() {
+		assertThrowsExactly(IllegalArgumentException.class, () -> userService.verifyEmail("", userId), "code is required; it must not be null");
```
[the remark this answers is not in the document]
CanadianPhilipBarkow: Implemented in commit 70d88ff21299e8b6f77785dd0b22d60b10f1a733.

backend/src/test/java/ca/gov/dtsstn/cdcp/api/service/UserServiceTests.java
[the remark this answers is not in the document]
CanadianPhilipBarkow: As of commit 70d88ff21299e8b6f77785dd0b22d60b10f1a733 I have attempted to apply the test comments generally, where applicable.

backend/src/test/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersControllerIT.java
```diff
@@ -106,4 +108,53 @@ void testGetUserById_Forbidden() throws Exception {
 			.andExpect(status().isForbidden());
 	}
 
+	@Test
+	@DisplayName("Test authenticated POST /api/v1/users/{userid}/email-validations with valid code")
```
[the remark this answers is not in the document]
CanadianPhilipBarkow: Punctuation corrected in commit 70d88ff21299e8b6f77785dd0b22d60b10f1a733.

backend/src/test/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersControllerIT.java
```diff
@@ -106,4 +108,53 @@ void testGetUserById_Forbidden() throws Exception {
...
+	@WithMockUser(roles = { "Users.Administer" })
+	void testVerifyConfirmationCodeStatus_Valid() throws Exception {
+		final var mockUser = ImmutableUser.builder()
+			.id("00000000-0000-0000-0000-000000000000")
+			.email("user@example.com")
+			.emailVerified(false)
+			.addUserAttributes(ImmutableUserAttribute.builder()
+				.name("EXAMPLE_ATTRIBUTE")
+				.value("42")
```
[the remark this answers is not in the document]
CanadianPhilipBarkow: Implemented in commit 70d88ff21299e8b6f77785dd0b22d60b10f1a733.

backend/src/test/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersControllerIT.java
```diff
@@ -106,4 +108,53 @@ void testGetUserById_Forbidden() throws Exception {
...
+				.build())
```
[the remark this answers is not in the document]
CanadianPhilipBarkow: Implemented in commit 70d88ff21299e8b6f77785dd0b22d60b10f1a733.

backend/src/test/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersControllerIT.java
```diff
@@ -106,4 +108,53 @@ void testGetUserById_Forbidden() throws Exception {
...
 	}
 
+	@Test
+	@DisplayName("Test authenticated POST /api/v1/users/{userid}/email-validations with valid code")
+	@WithMockUser(roles = { "Users.Administer" })
+	void testVerifyConfirmationCodeStatus_Valid() throws Exception {
+		final var mockUser = ImmutableUser.builder()
+			.id("00000000-0000-0000-0000-000000000000")
+			.email("user@example.com")
+			.emailVerified(false)
```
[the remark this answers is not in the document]
CanadianPhilipBarkow: Sure, implemented in commit 70d88ff21299e8b6f77785dd0b22d60b10f1a733.

backend/src/test/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersControllerIT.java
CanadianPhilipBarkow: Both this test and endpoint in the controller have been updated as of commit 70d88ff21299e8b6f77785dd0b22d60b10f1a733.

backend/src/test/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersControllerIT.java
CanadianPhilipBarkow: I have attempted to generalize the comments as of commit 70d88ff21299e8b6f77785dd0b22d60b10f1a733.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
nicholas-ly: Although `ResourceNotFoundException` is thrown, it is currently thrown at the service level. 

The exception should be thrown at the web/controller level instead as it is the controller's responsibility to handle web requests/responses (ie. returning 404 to the consumer). Take a look at `UsersController` and/or `SubscriptionsController` to see how that exception is currently being thrown.

backend/src/test/java/ca/gov/dtsstn/cdcp/api/service/UserServiceTests.java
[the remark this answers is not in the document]
nicholas-ly: FYI the inline instantiation isn't in place for `testVerifyEmail_Valid()`.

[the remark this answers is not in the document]
gregory-j-baker: > I tend to think that inline object instantiation makes it harder to follow what's happening

I tend to agree, but only in certain cases. Here, the builder pattern makes it easy to understand that you are setting the `createdDate`, `expiryDate`, etc.. so the pattern is effectively self-documenting.

In cases where you are doing something like this: `new FooBar("Coder McCodeface", "Head reviewer", false, true, true, WAY_BELOW_AVERAGE)` then it does make sense to move the parameters into separate instances.

But compare the above line to:

``` java
Foobar.builder()
  .name("Coder McCodeface")
  .title("Head reviewer")
  .isAdmired(false)
  .isOverworked(true)
  .lovesReviewingPRs(true)
  .payGrade(WAY_BELOW_AVERAGE)
  .build()
```

The builder pattern makes it easy to see what each value represents.

PR #1530, opened 2024-05-29
Add service-level bean validation

gregory-j-baker:
### Add bean validation to service classes

This PR adds bean validation to the `UserService` class and defines different validation schemes for creating and updating data.


gregory-j-baker:
Closing.. will open new PR with similar changes.

PR #1684, opened 2024-06-10
4059 SYSTEST - The application information page is missing error message when input same sin as child

Fbasham:
### Description
More zod fun...
What's happening with this issue is that the validation was happening successfully, but because `path` was specified as an attribute in the custom error (and `superRefine` is being applied on a field level and not on the form object level), it was being added as a nested object in the returned errors after zod does it's validation. This messes with how we extract the error because we only expect flat objects.  

I ended up removing unnecessary attributes and returns, and standardizing the SIN validation across all of the flows.  The errors should successfully display now. 

[AB#4059](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/4059)


shawn320 (approved):
Tested and functioning as expected!

PR #1706, opened 2024-06-12
AB#3369 - backend user json patch

leiye29:
### Description

backend user json patch

[AB#3369](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/3369)

image

### Checklist

### DIFF
--- a/backend/src/main/java/ca/gov/dtsstn/cdcp/api/service/UserService.java
+++ b/backend/src/main/java/ca/gov/dtsstn/cdcp/api/service/UserService.java
@@ -5,6 +5,7 @@ import java.util.Optional;
 import java.util.function.Predicate;
 
 import org.apache.commons.lang3.RandomStringUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.mapstruct.factory.Mappers;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -134,18 +135,16 @@ public class UserService {
 		return userRepository.findById(id).map(userMapper::toUser);
 	}
 
-	public void updateUser(String userId, String email) {
+	public void updateUser(String userId, User userPatch) {
 		Assert.hasText(userId, "userId is required; it must not be null or blank");
-		Assert.hasText(email, "email is required; it must not be null or blank");
+		Assert.notNull(userPatch, "userPatch is required; it must not be null");
 
 		log.debug("Fetching user [{}] from repository", userId);
 		final var user = userRepository.findById(userId).orElseThrow();
-
-		if (email.equals(user.getEmail()) == false) {
-			user.setEmail(email);
+		if (StringUtils.equals(userPatch.getEmail(), user.getEmail()) == false) {
 			user.setEmailVerified(false);
 
-			userRepository.save(user);
+			userRepository.save(userMapper.update(user, userPatch));
 		}
 	}
 
--- a/backend/src/main/java/ca/gov/dtsstn/cdcp/api/service/UserService.java
+++ b/backend/src/main/java/ca/gov/dtsstn/cdcp/api/service/UserService.java
@@ -141,11 +141,12 @@ public class UserService {
 
 		log.debug("Fetching user [{}] from repository", userId);
 		final var user = userRepository.findById(userId).orElseThrow();
+
 		if (StringUtils.equals(userPatch.getEmail(), user.getEmail()) == false) {
 			user.setEmailVerified(false);
-
-			userRepository.save(userMapper.update(user, userPatch));
 		}
+
+		userRepository.save(userMapper.update(user, userPatch));
 	}
 
 	public void updateSubscriptionForUser(String userId, String subscriptionId, String languageId) {
--- a/backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
+++ b/backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
@@ -2,6 +2,8 @@ package ca.gov.dtsstn.cdcp.api.web.v1.controller;
 
 import org.mapstruct.factory.Mappers;
 import org.springframework.util.Assert;
+import org.springframework.validation.BindException;
+import org.springframework.validation.BindingResult;
 import org.springframework.validation.annotation.Validated;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PathVariable;
@@ -9,16 +11,20 @@ import org.springframework.web.bind.annotation.PatchMapping;
 import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
+
 import ca.gov.dtsstn.cdcp.api.config.SpringDocConfig.OAuthSecurityRequirement;
 import ca.gov.dtsstn.cdcp.api.service.UserService;
 import ca.gov.dtsstn.cdcp.api.web.exception.ResourceNotFoundException;
+import ca.gov.dtsstn.cdcp.api.web.json.JsonPatchProcessor;
 import ca.gov.dtsstn.cdcp.api.web.v1.model.UserModel;
-import ca.gov.dtsstn.cdcp.api.web.v1.model.UserUpdateModel;
 import ca.gov.dtsstn.cdcp.api.web.v1.model.mapper.UserModelMapper;
 import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.Parameter;
 import io.swagger.v3.oas.annotations.responses.ApiResponse;
 import io.swagger.v3.oas.annotations.tags.Tag;
+
+import jakarta.json.JsonPatch;
+import jakarta.validation.ConstraintViolationException;
 import jakarta.validation.constraints.NotBlank;
 
 @Validated
@@ -32,9 +38,13 @@ public class UsersController {
 
 	private final UserService userService;
 
-	public UsersController(UserService userService) {
+	private final JsonPatchProcessor jsonPatchProcessor;
+
+	public UsersController(UserService userService, JsonPatchProcessor jsonPatchProcessor) {
 		Assert.notNull(userService, "userService is required; it must not be null");
+		Assert.notNull(jsonPatchProcessor, "jsonPatchProcessor is required; it must not be null");
 		this.userService = userService;
+		this.jsonPatchProcessor = jsonPatchProcessor;
 	}
 
 	@GetMapping({ "/{id}" })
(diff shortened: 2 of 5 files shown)


backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/util/Functions.java
```diff
@@ -0,0 +1,26 @@
...
+import com.github.fge.jsonpatch.JsonPatch;
+import com.github.fge.jsonpatch.JsonPatchException;
+
+
+@Component
+public class Functions {
+
+	public <T> T applyJsonPatch(JsonPatch patch, T targetModel, Class<T> targetModelClass) {
+		try {
+			final ObjectMapper objectMapper = new ObjectMapper();
```
gregory-j-baker, line 19: I feel like this should be a class field, since instantiating it is *very* expensive.

``` java
private final ObjectMapper objectMapper = new ObjectMapper()
    .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
    .findAndRegisterModules();
```

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/util/Functions.java
```diff
@@ -0,0 +1,26 @@
...
+			JsonNode patched = patch.apply(objectMapper.convertValue(targetModel, JsonNode.class));
+			return objectMapper.treeToValue(patched, targetModelClass);
```
gregory-j-baker, line 21: I'm no 100% sure, but it feels like there's a potential `NullPointerException` hiding in here somewhere. You should do some assertions on the incoming parameters so you can at least control the error message that is thrown if they are null.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/model/UserPatchModel.java
```diff
@@ -9,8 +9,8 @@
 
 @Immutable
 @Schema(name = "UserUpdate")
```
gregory-j-baker, line 11: This should be updated to match the model name.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
```diff
@@ -48,20 +60,26 @@ public UserModel getUserById(
...
 				.orElseThrow(() -> new ResourceNotFoundException("No user with id=[%s] was found".formatted(id)));
 
-		userService.updateUser(user.getId(), userUpdateModel.getEmail());
-	}	
+		final var userPatchModel = userPatchModelMapper.toModel(user);
+		final var userPatched = functions.applyJsonPatch(patch, userPatchModel, UserPatchModel.class);
+
+		if (EmailValidator.getInstance().isValid(userPatched.getEmail()) == false) {
+			throw new ResourceNotFoundException("Invalid email=[%s]".formatted(userPatched.getEmail()));
+		}
```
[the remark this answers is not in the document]
gregory-j-baker: Also, `ResourceNotFoundException` isn't the correct error to throw here. This will result in a 404 error, but it *should* be a 400 error.

I sent you a message in Slack about a different way to implement validation, if you decide to go in that new direction, then a 400 response will automatically occur when validation fails.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/util/Functions.java
```diff
@@ -0,0 +1,26 @@
...
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.github.fge.jsonpatch.JsonPatch;
+import com.github.fge.jsonpatch.JsonPatchException;
+
+
+@Component
+public class Functions {
```
gregory-j-baker, line 15: I sent you a message in Slack about a different (in my opinion, more robust) way to implement this.

gregory-j-baker (changes requested):

gregory-j-baker (changes requested):
I also just tried the following patch in your branch and it failed:

``` json
[{
  "op": "remove",
  "path": "/email"
}]
```

The error was:

``` json
{
  "timestamp": "2024-06-12T11:42:29.757+00:00",
  "status": 400,
  "error": "Bad Request",
  "message": "JSON Patch fails",
  "path": "/api/v1/users/fade01ac-097c-4f14-99dc-46155b863305"
}
```

There are two issues here:
- I don't think this should have failed
- When a validation failure does occur, we should be providing the reason why so the consumer can correct the request on their end

I believe that the implementation that I suggested in Slack will return good error messages when validation fails (although I didn't test that).

backend/pom.xml
```diff
@@ -107,6 +109,16 @@
...
+		<dependency>
+			<groupId>com.github.java-json-tools</groupId>
+			<artifactId>json-patch</artifactId>
+			<version>${json-patch.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>commons-validator</groupId>
+			<artifactId>commons-validator</artifactId>
+			<version>${commons-validator.version}</version>
+		</dependency>
```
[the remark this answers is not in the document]
leiye29: removed

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/util/Functions.java
leiye29: added

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/model/UserPatchModel.java
leiye29: fixed

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/util/Functions.java
gregory-j-baker: It should be a class field, not a method-scoped variable.

Note that this whole class should be removed when #1710 is merged.

leiye29: removed after #1710 was merged.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/util/Functions.java
leiye29: removed after #1701 was merged

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/util/Functions.java
leiye29: removed after #1701 was merged

backend/src/main/java/ca/gov/dtsstn/cdcp/api/service/UserService.java
```diff
@@ -134,18 +135,16 @@ public Optional<User> getUserById(String id) {
 		return userRepository.findById(id).map(userMapper::toUser);
 	}
 
-	public void updateUser(String userId, String email) {
+	public void updateUser(String userId, UserPatch userPatched) {
 		Assert.hasText(userId, "userId is required; it must not be null or blank");
-		Assert.hasText(email, "email is required; it must not be null or blank");
+		Assert.notNull(userPatched, "patchedUser is required; it must not be null");
```
gregory-j-baker, line 140:
```suggestion
	public void updateUser(String userId, UserPatch userPatch) {
		Assert.hasText(userId, "userId is required; it must not be null or blank");
		Assert.notNull(userPatch, "userPatch is required; it must not be null");
```

backend/src/main/java/ca/gov/dtsstn/cdcp/api/service/UserService.java
```diff
@@ -134,18 +135,16 @@ public Optional<User> getUserById(String id) {
...
 
 		log.debug("Fetching user [{}] from repository", userId);
 		final var user = userRepository.findById(userId).orElseThrow();
-
-		if (email.equals(user.getEmail()) == false) {
-			user.setEmail(email);
+		if (userPatched.getEmail().equals(user.getEmail()) == false) {
```
gregory-j-baker, line 144: Since either of these values can be null, you should use `StringUtils.equals(..)` from `org.apache.commons.lang3` to compare.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/service/domain/UserPatch.java
```diff
@@ -0,0 +1,13 @@
+package ca.gov.dtsstn.cdcp.api.service.domain;
+
+import org.immutables.value.Value.Immutable;
+
+import jakarta.annotation.Nullable;
+
+@Immutable
+public interface UserPatch extends BaseDomainObject {
```
gregory-j-baker, line 8: I don't think this object should subclass `BaseDomainObject`.

When you are updating a stored object, the fields within `BaseDomainObject` should be readonly. It doesn't make sense to allow for them to be set. Just remove the `extends BaseDomainObject` and you should be okay.

backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/UsersController.java
```diff
@@ -48,20 +55,22 @@ public UserModel getUserById(
...
 
 		final var user = userService.getUserById(id)
 				.orElseThrow(() -> new ResourceNotFoundException("No user with id=[%s] was found".formatted(id)));
 
-		userService.updateUser(user.getId(), userUpdateModel.getEmail());
-	}	
+		final var userPatchModel = userModelMapper.toPatchModel(user);
+		final var userPatched = jsonPatchProcessor.patch(userPatchModel, patch);
 
+		userService.updateUser(id, userModelMapper.toPatchDomain(userPatched));
```
gregory-j-baker, line 73: I wonder if it is even necessary to have a patch domain object here? Couldn't you just do something like:

``` java
final var user = userService.getUserById(id)
    .orElseThrow(() -> new ResourceNotFoundException("No user with id=[%s] was found".formatted(id)));

final var userModel = userModelMapper.toModel(user);
final var userModelPatched = jsonPatchProcessor.patch(userModel, patch);

userService.updateUser(id, userModelMapper.toDomain(userModelPatched));
```

gregory-j-baker: Also.. to correctly handle invalid data, you can do this:

``` java
try {
    final var userModelPatched = jsonPatchProcessor.patch(userModel, patch);  
    userService.updateUser(id, userModelMapper.toDomain(userModelPatched)); 
}
catch (final ConstraintViolationException constraintViolationException) {
    final var bindException = new BindException(jsonPatch, "jsonPatch");

    constraintViolationException.getConstraintViolations().forEach(constraintViolation -> {
        final var errorCode = constraintViolation.getPropertyPath().toString();
        final var message = constraintViolation.getMessage();
        bindException.reject(errorCode, message);
    });

    throw bindException;
}
```

gregory-j-baker (changes requested):

backend/src/main/java/ca/gov/dtsstn/cdcp/api/service/UserService.java
gregory-j-baker: In case it isn't clear.. The `Assert.notNull(userPatched, "patchedUser is required; it must not be null")` line had a typo or copy/paste error.. while suggesting a fix for that I also renamed the variable from `userPatched` to `userPatch`.

leiye29: fixed

backend/src/main/java/ca/gov/dtsstn/cdcp/api/service/UserService.java
leiye29: fixed

backend/src/main/java/ca/gov/dtsstn/cdcp/api/service/domain/UserPatch.java
leiye29: this class is removed

gregory-j-baker (approved):

PR #1718, opened 2024-06-13
use InputSanitizeField in place of InputField

Fbasham:
### Description
Swap out the `InputField` with `InputSanitizeField` on name fields across all flows to match server-side validation on characters put in place yesterday.


sebastien-comeau (approved):
Great! I wonder if we have to do the same for other fields like addresses.

PR #1726, opened 2024-06-13
Add lookup server utils to unify localization and sorting regions lookup

sebastien-comeau:
### Description

Add lookup server utils to unify localization and sorting regions lookup within the application.

### Checklist

- [x] I have tested the changes locally
- [x] I have updated the documentation if necessary
- [x] I have added/updated tests that prove my fix is effective or that my feature works
- [x] I have checked that my code follows the project's coding style by running `npm run format:check`
- [x] I have checked that my code contains no linting errors by running `npm run lint`
- [x] I have checked that my code contains no type errors by running `npm run typecheck`
- [x] I have checked that all unit tests pass by running `npm run test:unit -- run`
- [x] I have checked that all e2e tests pass by running `npm run test:e2e`

PR #1731, opened 2024-06-14
order preferred languages by locale

Fbasham:
### Description
Based on Figma, the preferred communication language for a user should be sorted based on their locale. Now, since there are 2 official languages in Canada, the way they have it presented is that if the locale of the application is in French, then `Francais` should come before `Anglais`.  If the locale of the application is English, then `English`comes before `French`.

This could be probably be moved into the newly created `lookup-utils`, but for now I'm holding back until that stabilizes.

https://dev.azure.com/ESDCCM/CDCP/_workitems/edit/9003

image

image

### Checklist

### DIFF
--- a/frontend/app/utils/lookup-utils.server.ts
+++ b/frontend/app/utils/lookup-utils.server.ts
@@ -1,5 +1,5 @@
 import { getEnv } from './env.server';
-import { Country, MaritalStatus, Region } from '~/services/lookup-service.server';
+import { Country, Language, MaritalStatus, Region } from '~/services/lookup-service.server';
 
 /**
  * Localizes a single country object by adding a localized name.
@@ -117,4 +117,36 @@ export function localizeRegions(regions: ReadonlyArray<Region>, locale: AppLocal
  */
 export function localizeAndSortRegions(regions: ReadonlyArray<Region>, locale: AppLocale) {
   return localizeRegions(regions, locale).toSorted((a, b) => a.name.localeCompare(b.name, locale));
+}
+
+/**
+ * Localizes a single language object by adding a localized name.
+ *
+ * @param language - The language object to localize.
+ * @param locale - The locale code for localization.
+ * @returns The localized language object with a localized name.
+ */
+export function localizeLanguage(language: Language, locale: string) {
+  const { nameEn, nameFr, ...rest } = language;
+  return {
+    ...rest,
+    name: locale === 'fr' ? nameFr : nameEn,
+  };
+}
+
+/**
+ * Localizes an array of language objects by adding localized names and sorting them.
+ *
+ * @param language - The array of language objects to localize.
+ * @param locale - The locale code for localization.
+ * @param firstLanguageId - The language ID that specifies the language object that should appear first in the sorted array.
+ * @returns The localized and sorted array of language objects.
+ */
+export function localizeAndSortPreferredLanguages(languages: Language[], locale: string, firstLanguageId?: number) {
+  const mappedLanguages = languages.map((language) => localizeLanguage(language, locale));
+  return mappedLanguages.toSorted((a, b) => {
+    if (firstLanguageId && a.id === firstLanguageId.toString()) return -1;
+    if (firstLanguageId && b.id === firstLanguageId.toString()) return 1;
+    return a.name.localeCompare(b.name, locale);
+  });
 }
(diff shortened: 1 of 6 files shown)


gregory-j-baker:
> This sorting logic should be handle in one place to unify it within the application, similar to #1726
> 
> I think what you're trying to do is a bit hard to understand and not sure if it would work correctly if we had more than 2 languages.
> 
> If what you're trying do to is to move the current locale's language as first element and sort the rest then this would be a better way.
> 
> ```js
> // Sample array of objects
> const array = [
>   { id: 3, name: 'Charlie' },
>   { id: 1, name: 'Alice' },
>   { id: 4, name: 'David' },
>   { id: 2, name: 'Bob' }
> ];
> 
> // The id of the element to move to the top
> const idToMove = 4; // Change this to an id that exists in the array
> 
> // Step 1: Find the element with the specific id
> const elementToMove = array.find(item => item.id === idToMove);
> 
> // Step 2: Filter out the element to move from the original array
> const filteredArray = array.filter(item => item.id !== idToMove);
> 
> // Step 3: Sort the remaining array by the 'name' property using toSorted with the 'en' locale
> const sortedArray = filteredArray.toSorted((a, b) => a.name.localeCompare(b.name, 'en'));
> 
> // Step 4: Create the new array with the element moved to the top if it exists
> const newArray = elementToMove ? [elementToMove, ...sortedArray] : sortedArray;
> 
> console.log(newArray);
> console.log(array); // Original array remains unchanged
> ```

The way that Frank is doing it is pretty conventional, in my opinion. The implementation is just uglified because of 1) the data structures that are being used, and 2) using a ternary instead of more self-documenting code.

gregory-j-baker:
> ```js
> // some code
> ```

A (in my opinion) cleaner implementation:

``` javascript
const array = [
  { id: 3, name: "Charlie" },
  { id: 1, name: "Alice" },
  { id: 4, name: "David" },
  { id: 2, name: "Bob" },
];

function withCurrentItemFirst(itemId) {
  return (a, b) => {
    if (a.id === itemId) return -1;
    if (b.id === itemId) return 1;
    return a.name.localeCompare(b.name);
  };
}

const sortedArray = [...array].sort(withCurrentItemFirst(4));
console.log({ array, sortedArray });
```

frontend/app/utils/lookup-utils.server.ts
```diff
@@ -27,3 +27,34 @@ export function localizeAndSortRegions(regions: Region[], locale: string) {
...
+ *
+ * @param language - The language object to localize.
+ * @param locale - The locale code for localization.
+ * @returns The localized language object with a localized name.
+ */
+export function localizeLanguage(language: Language, locale: string) {
+  const { nameEn, nameFr, ...rest } = language;
+  return {
+    ...rest,
+    name: (locale === 'fr' ? nameFr : nameEn) ?? '',
```
sebastien-comeau, line 42: If it `nameFr` or `nameEn` is `undefined` and let it be. I think we should throw an exception if preferred language names are `undefined`.

frontend/app/utils/lookup-utils.server.ts
```diff
@@ -27,3 +27,34 @@ export function localizeAndSortRegions(regions: Region[], locale: string) {
...
+ * @param locale - The locale code for localization.
+ * @param firstLanguageId - The language ID that specifies the language object that should appear first in the sorted array.
+ * @returns The localized and sorted array of language objects.
+ */
+export function localizeAndSortPreferredLanguages(languages: Language[], locale: string, firstLanguageId: number) {
+  const mappedLanguages = languages.map((language) => localizeLanguage(language, locale));
+  const firstLanguage = mappedLanguages.find((language) => language.id === firstLanguageId.toString());
+  const filteredLanguages = mappedLanguages.filter((language) => language.id !== firstLanguageId.toString());
+  const sortedLanguages = filteredLanguages.toSorted((a, b) => a.name.localeCompare(b.name, locale));
+  return firstLanguage ? [firstLanguage, ...sortedLanguages] : sortedLanguages;
```
sebastien-comeau, line 59: Can be improved as suggested by @gregory-j-baker

```suggestion
  return mappedLanguages.toSorted((a, b) => {
    if (firstLanguageId && a.id === firstLanguageId.toString()) return -1;
    if (firstLanguageId && b.id === firstLanguageId.toString()) return 1;
    return a.name.localeCompare(b.name, locale);
  });
```

sebastien-comeau (changes requested):

frontend/app/utils/lookup-utils.server.ts
sebastien-comeau: Set the `firstLanguageId` param optional.

```
firstLanguageId?: number
```

sebastien-comeau (dismissed):
LGTM

sebastien-comeau (approved):

PR #1755, opened 2024-06-17
A11y - Using Placeholder text as a label

Fbasham:
### Description
Adds DD/JJ for days, and YYYY/AAAA for years in the labels for the date picker. This was flagged by the accessibility office.

AB#4109


sebastien-comeau (changes requested):
Remove the placeholder to Day and Year inputs.

frontend/app/components/date-picker-field.tsx
```diff
@@ -311,7 +309,7 @@ interface DatePickerDayProps {
   id: string;
   label: string;
   name: string;
-  placeholder: string;
+  placeholder?: string;
```
sebastien-comeau, line 312: Remove `placeholder` prop since it's not used.

frontend/app/components/date-picker-field.tsx
```diff
@@ -267,7 +265,7 @@ interface DatePickerYearProps {
...
```
sebastien-comeau, line 268: Remove `placeholder` prop since it's not used.

sebastien-comeau (changes requested):

frontend/app/components/date-picker-field.tsx
Fbasham: it's not used in this component...doesn't mean it's never not going to be used though

Fbasham: I'm not going to die on this hill, but I don't necessarily agree either.  Changes made.

sebastien-comeau: There's a blog posts and articles on that placeholder is bad for user's experience and I agree.

sebastien-comeau (approved):

frontend/app/components/date-picker-field.tsx
sebastien-comeau: https://www.nngroup.com/articles/form-design-placeholders/

PR #1820, opened 2024-06-21
4152 add localize prov/fed/terr benefits to look-up service

Fbasham:
### Description
This PR moves the (duplicated) logic of localizing, filtering, and joining federal/provincial/territorial benefits on the review-information screens to the lookup-utils.  

The rationale for keeping a combination of `filter/map/join` instead of `find` is that in my mind, it may be possible in the future to select multiple benefits as a checkbox instead of radio buttons.  

[AB#4152](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/4152)

### Checklist

### DIFF
--- a/frontend/app/routes/$lang/_public/apply/$id/child/review-child-information.tsx
+++ b/frontend/app/routes/$lang/_public/apply/$id/child/review-child-information.tsx
@@ -22,9 +22,10 @@ import { getLookupService } from '~/services/lookup-service.server';
 import { parseDateString, toLocaleDateString } from '~/utils/date-utils';
 import { getEnv } from '~/utils/env.server';
 import { useHCaptcha } from '~/utils/hcaptcha-utils';
-import { getNameByLanguage, getTypedI18nNamespaces } from '~/utils/locale-utils';
+import { getTypedI18nNamespaces } from '~/utils/locale-utils';
 import { getFixedT, getLocale } from '~/utils/locale-utils.server';
 import { getLogger } from '~/utils/logging.server';
+import { localizeFederalSocialProgram, localizeProvincialTerritorialSocialProgram } from '~/utils/lookup-utils.server';
 import { mergeMeta } from '~/utils/meta-utils';
 import { RouteHandleData, getPathById } from '~/utils/route-utils';
 import { getTitleMetaTags } from '~/utils/seo-utils';
@@ -56,15 +57,16 @@ export async function loader({ context: { session }, params, request }: LoaderFu
   const locale = getLocale(request);
   const lookupService = getLookupService();
 
-  const allFederalSocialPrograms = lookupService.getAllFederalSocialPrograms();
-  const allProvincialTerritorialSocialPrograms = lookupService.getAllProvincialTerritorialSocialPrograms();
-
   const hCaptchaEnabled = ENABLED_FEATURES.includes('hcaptcha');
 
   const csrfToken = String(session.get('csrfToken'));
   const meta = { title: t('gcweb:meta.title.template', { title: t('apply-child:review-child-information.page-title') }) };
 
   const children = state.children.map((child) => {
+    const selectedFederalBenefit = child.dentalBenefits.federalSocialProgram && localizeFederalSocialProgram(lookupService.getFederalSocialProgramById(child.dentalBenefits.federalSocialProgram), locale);
+    const selectedProvincialBenefit =
+      child.dentalBenefits.provincialTerritorialSocialProgram && localizeProvincialTerritorialSocialProgram(lookupService.getProvincialTerritorialSocialProgramById(child.dentalBenefits.provincialTerritorialSocialProgram), locale);
+
     return {
       id: child.id,
       firstName: child.information.firstName,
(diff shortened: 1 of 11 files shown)


frontend/app/utils/lookup-utils.server.ts
```diff
@@ -150,3 +150,62 @@ export function localizeAndSortPreferredLanguages(languages: Language[], locale:
...
+}
+
+/**
+ *
+ * @param programs - The array of federal social program objects to localize and filter
+ * @param locale - The locale code for localization
+ * @param targetProgramId - The federal social program ID to filter on
+ * @returns - The localized, filtered, and joined array of federal social programs by target federal social program
+ */
+export function localizeFilterJoinFederalSocialPrograms(programs: FederalSocialProgram[], locale: string, targetProgramId: string | undefined) {
```
sebastien-comeau, line 176: I believe the filtering or search operation should be handled at the service level. Filtering by the primary key doesn't make sense here, as it will either return an empty array or an array with a single item, making the array join irrelevant.

I recommend creating a findFederalSocialProgram(id: string): FederalSocialProgram | null function that utilizes the getAllFederalSocialPrograms() function at the service level. If the result is not null, you can then call localizeFederalSocialProgram. If null is an invalid result, you can use invariant to throw an exception.

Don't forget to add `localizeFederalSocialPrograms` that will localize an array of `FederalSocialProgram` and also the `localizeAndSortFederalSocialPrograms` that will sort by locale as well. Those functions can be used in the edit screens.

sebastien-comeau (changes requested):

frontend/app/utils/lookup-utils.server.ts
sebastien-comeau: @Fbasham updated my comment ^

frontend/app/services/lookup-service.server.ts
```diff
@@ -376,6 +376,19 @@ function createLookupService() {
     return federalSocialPrograms;
   }
 
+  function getFederalSocialProgramsById(id: string) {
+    log.debug('Fetching federal social program with id: [%j]', id);
+
+    const federalSocialProgram = getAllFederalSocialPrograms().find((program) => program.id === id);
+
+    if (!federalSocialProgram) {
+      throw new Error('Failed to find federal social program');
```
sebastien-comeau, line 385:
```suggestion
      throw new Error(`Failed to find federal social program; id: ${id}`);
```

Same for `getProvincialTerritorialSocialProgramById`

frontend/app/services/lookup-service.server.ts
```diff
@@ -376,6 +376,19 @@ function createLookupService() {
     return federalSocialPrograms;
   }
 
+  function getFederalSocialProgramsById(id: string) {
+    log.debug('Fetching federal social program with id: [%j]', id);
```
sebastien-comeau, line 380: Use `%s` instead of `%j` since `id` is not a JSON/Object type.

Same for `getProvincialTerritorialSocialProgramById`

https://github.com/winstonjs/winston?tab=readme-ov-file#string-interpolation

https://nodejs.org/dist/latest/docs/api/util.html#utilformatformat-args
image

frontend/app/routes/$lang/_public/apply/$id/child/review-child-information.tsx
```diff
@@ -76,18 +78,12 @@ export async function loader({ context: { session }, params, request }: LoaderFu
         acessToDentalInsurance: child.dentalInsurance,
         federalBenefit: {
           access: child.dentalBenefits.hasFederalBenefits,
-          benefit: allFederalSocialPrograms
-            .filter((obj) => obj.id === child.dentalBenefits.federalSocialProgram)
-            .map((obj) => getNameByLanguage(locale, obj))
-            .join(', '),
+          benefit: selectedFederalBenefit ? selectedFederalBenefit.name : '',
```
sebastien-comeau, line 81: Use `undefined` when `selectedFederalBenefit ` is `undefined` else use `name`. Same logic in other places.

```suggestion
          benefit: selectedFederalBenefit && selectedFederalBenefit.name,
```

frontend/app/routes/$lang/_public/apply/$id/child/review-child-information.tsx
```diff
@@ -76,18 +78,12 @@ export async function loader({ context: { session }, params, request }: LoaderFu
...
         },
         provTerrBenefit: {
           access: child.dentalBenefits.hasProvincialTerritorialBenefits,
           province: child.dentalBenefits.province,
-          benefit: allProvincialTerritorialSocialPrograms
-            .filter((obj) => obj.id === child.dentalBenefits.provincialTerritorialSocialProgram)
-            .map((obj) => getNameByLanguage(locale, obj))
-            .join(', '),
+          benefit: selectedProvincialBenefit ? selectedProvincialBenefit.name : '',
```
sebastien-comeau, line 86: Use `undefined` when `selectedProvincialBenefit` is `undefined` else use `name`. Same logic in other places.
```suggestion
          benefit: selectedProvincialBenefit && selectedProvincialBenefit.name,
```

sebastien-comeau (changes requested):

sebastien-comeau (approved):

PR #1833, opened 2024-06-25
[gitops] Enable maintenance mode in production and deploy v2.0.0

gregory-j-baker:
Per title.


sebastien-comeau (approved):

sebastien-comeau:
Approved!

PR #1879, opened 2024-07-02
Returning 405 for GET `/users/{userId}/email-validations`

nicholas-ly:
### Description
...and some refactoring of annotation attributes.

GET `/users/{userId}/email-validations` was originally exposed to create a HATEOAS `link` from the user resource and would return a 200 with an empty collection. The response has been changed to return a 405 - Method not allowed, which is more suitable.

Also removed `required` attribute from `@Parameter` because `@PathVariable`s are required by default.

**Before:**
image

**After:**
image

### Test Instructions
Run application locally and try a request to GET `/users/{userId}/email-validations`. A 405 should be returned.

### DIFF
--- a/backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/EmailValidationsController.java
+++ b/backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/EmailValidationsController.java
@@ -1,10 +1,7 @@
 package ca.gov.dtsstn.cdcp.api.web.v1.controller;
 
-import static org.springframework.hateoas.server.mvc.WebMvcLinkBuilder.linkTo;
-import static org.springframework.hateoas.server.mvc.WebMvcLinkBuilder.methodOn;
-
-import org.springframework.hateoas.CollectionModel;
 import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
 import org.springframework.util.Assert;
 import org.springframework.validation.BindException;
 import org.springframework.validation.BindingResult;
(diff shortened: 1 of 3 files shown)


backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/controller/EmailValidationsController.java
```diff
@@ -43,24 +39,21 @@ public EmailValidationsController(UserService userService) {
 	}
 
 	@GetMapping
+	@ResponseStatus(HttpStatus.METHOD_NOT_ALLOWED)
 	@Operation(summary = "List all email validations for a user")
-	public CollectionModel<EmailValidationModel> getEmailValidationByUserId(
+	public ResponseEntity<Void> getEmailValidationByUserId(
```
sebastien-comeau, line 44: Why not removing this GET endpoint? Wouldn't Spring handle it and returns 405?

nicholas-ly: I agree and it was what I originally aimed to do, but the problem is the following to create a HATEOAS link:
https://github.com/DTS-STN/canadian-dental-care-plan/blob/58dd9075ded63fd4e165de224e94d4d6275e1263/backend/src/main/java/ca/gov/dtsstn/cdcp/api/web/v1/model/mapper/UserModelMapper.java#L47

Removing the endpoint would mean the mapper would have to `linkTo(..)` the POST endpoint - `verifyConfirmationCodeStatus(..)`.

The mapper would have to pass a `emailValidationModel`, and a `bindingResult` argument to be able to link to this method and handle the `BindException`, which was really ugly IMO.

sebastien-comeau: HATEOAS.....

sebastien-comeau (approved):

PR #1932, opened 2024-07-11
refactor(frontend): Add zod utility func to transform flattened error

sebastien-comeau:
### Description

This pull request introduces the `transformFlattenedError` utility function from Zod to transform flattened errors. Since our object schemas are only one level deep, using [.flatten()](https://zod.dev/ERROR_HANDLING?id=flattening-errors) is more convenient. The `fieldErrors` properties contain arrays of error strings (string[]), but we only use the first element of each array. This utility has been implemented in the status checker.

Usage example:

```ts
const flattenedError = {
  fieldErrors: {
    name: ["Name is required", "Name is too short"],
    age: ["Age must be a number"]
  }
};
const result = transformFlattenedError(flattenedError);
console.log(result);
// Output: { name: "Name is required", age: "Age must be a number" }
```

The next feature to be introduced is the `useErrorSummary` custom hook, which will leverage this change.

### Checklist

### DIFF
--- a/frontend/__tests__/utils/zod-utils.server.test.ts
+++ b/frontend/__tests__/utils/zod-utils.server.test.ts
@@ -1,0 +1,59 @@
+import { describe, expect, it } from 'vitest';
+
+import { transformFlattenedError } from '~/utils/zod-utils.server';
+
+describe('transformFlattenedError', () => {
+  it('should transform a flattened error object with multiple fields and multiple error messages', () => {
+    const flattenedError = {
+      fieldErrors: {
+        name: ['Name is required', 'Name is too short'],
+        age: ['Age must be a number'],
+      },
+    };
+    const expectedResult = {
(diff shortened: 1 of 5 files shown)


frontend/__tests__/utils/zod-utils.test.ts
```diff
@@ -0,0 +1,59 @@
+import { describe, expect, it } from 'vitest';
+
+import { transformFlattenedError } from '~/utils/zod.utils';
```
nicholas-ly, line 3: Since the Zod utils is only used by `action(..)`/server-side and to follow existing naming util file naming conventions:

```suggestion
import { transformFlattenedError } from '~/utils/zod-utils.server';
```

sebastien-comeau: Fair point. Fixed

PR #1946, opened 2024-07-16
AB#4216 -- Adding `docker-compose.yml` to spin up a Postgres instance

nicholas-ly:
### Description
As per title.

AB#4216

### Test Instructions
Navigate to `backend/docker` directory and run `docker-compose up`


gregory-j-baker (changes requested):
The new standard is to use a `compose.yaml` file instead of a `docker-compose.yaml` file.

<https://docs.docker.com/compose/compose-application-model/#the-compose-file>

You can also now omit the `version: '3'` line, since the compose plugin is smart enough to figure out what version you need.

gregory-j-baker:
> ## Test Instructions
> Navigate to backend/docker directory and run docker-compose up

Makes no difference to this PR, but the new way to do this is to use the compose plugin for docker (ie: `docker compose up`). <https://docs.docker.com/compose/install/>

The compose plugin is a native CLI plugin for the `docker` command, while `docker-compose` is a deprecated python script.

nicholas-ly:
> The new standard is to use a `compose.yaml` file instead of a `docker-compose.yaml` file.
> 
> https://docs.docker.com/compose/compose-application-model/#the-compose-file
> 
> You can also now omit the `version: '3'` line, since the compose plugin is smart enough to figure out what version you need.

Renamed to `compose.yaml` and removed `version: '3'` line.

gregory-j-baker (approved):

PR #1980, opened 2024-07-24
small refactor of getAlertType in status checker

Fbasham:
### Description
moves `getAlertType` into a utility file.  Previously this function was operating on a global variable which has since been changed to accept the `statusId` as the only argument.  

### Checklist

### DIFF
--- a/frontend/app/utils/application-code-utils.ts
+++ b/frontend/app/utils/application-code-utils.ts
@@ -1,3 +1,5 @@
+import { getEnv } from './env.server';
+
 export const applicationCodeInputPatternFormat = '### ### ### ####';
 const applicationCodeFormatRegex = /^\d{3}[ ]?\d{3}$/;
 const clientNumberFormatRegex = /^\d{3}[ ]?\d{3}[ ]?\d{3}[ ]?\d{2}$/;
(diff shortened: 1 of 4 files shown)


frontend/app/utils/application-code-utils.ts
```diff
@@ -55,3 +57,15 @@ export function formatSubmissionApplicationCode(applicationCode: string): string
   if (!/^\d{13}$/.test(strippedCode)) return applicationCode;
   return (strippedCode.match(/....$|.../g) ?? []).join(' ');
 }
+
+/**
+ *
+ * @param statusId - the statusId of the application to check
+ * @returns the "info" | "success" | "danger" which is consumed by the ContextualAlert component
+ */
+export function getContextualAlertType(statusId: string | undefined) {
```
nibivi77, line 66: If `getStatusIdWithSin` and/or `getStatusIdWithoutSin` do not find a statusId, does it return undefined? or null? 

If null, we might want to do this: 

`getContextualAlertType(statusId: string | undefined | null)`

Fbasham: good point.  I think it's probably better to operate on `statusId` (which will always be a string or null) instead of the `clientFriendlyStatus?.id` that I was before.

nicholas-ly (approved):

frontend/app/utils/application-code-utils.ts
Fbasham: resolving

PR #2009, opened 2024-07-31
Migrate to Vite!

gregory-j-baker:
### Migrate to Vite!

[Vite](https://vitejs.dev/) will eventually become the default compiler for Remix. To ensure that we have an upgrade path going forward, it is in our best interest to migrate.

## Description of changes

- removal of `express.server.ts` (functionality now moved to `entry.server.tsx` and various middlewares)
- updated `dockerfile` (now uses the vite build output)
- removal of the no-unused packages in `package.json`
- everything from the [Remix Vite migration guide](https://remix.run/docs/en/main/guides/vite#migrating)

### Checklist

- [x] I have tested the changes locally
- [x] I have added/updated tests that prove my fix is effective or that my feature works
- [x] I have checked that my code follows the project's coding style by running `npm run format:check`
- [x] I have checked that my code contains no linting errors by running `npm run lint`
- [x] I have checked that my code contains no type errors by running `npm run typecheck`
- [x] I have checked that all unit tests pass by running `npm run test:unit -- run`
- [x] I have checked that all e2e tests pass by running `npm run test:e2e`

PR #2012, opened 2024-08-01
Rename `env-utils.ts` → `env-utils.client.ts`

gregory-j-baker:
Since `env-utils.ts` references `window.env`, it should only be used client-side. Any server-side code that needs access to the environment can use `env.server.ts`.

This PR is to support migrating to vite, PR #2009, to make that PR a little smaller.

### Checklist


sebastien-comeau (commented):
Have you tested it? You need to be cautious with client-only code, as the code within this file can only be executed on the client side, not the server side. Even if it's within a page React component, that code might still be called on the server for SSR (Server-Side Rendering).

Run the application and go to http://localhost:3000/en/home.

image

This is the line that cause that issue since NavigationMenu is executed server-side (SSR) and client-side: https://github.com/DTS-STN/canadian-dental-care-plan/blob/44e582932767480edb3b37144f65f480a435233a/frontend/app/components/layouts/protected-layout.tsx#L60

gregory-j-baker:
I assumed that because the e2e tests passed, we were good. What's the point of having e2e tests if they can't detect something as trivial as your comment?

Anyway, I'll take a look.

sebastien-comeau:
> I assumed that because the e2e tests passed, we were good. What's the point of having e2e tests if they can't detect something as trivial as your comment?
> 
> Anyway, I'll take a look.

I get your point but the page load even though it returns a 500. 🤷

gregory-j-baker:
Closing this PR.. it is no longer necessary thanks to #2082.

PR #2076, opened 2024-08-20
playwright teamcity reporter

shawn320:
### Description
Added `playwright-teamcity-reporter` package into our playwright test configuration. 
I'm not sure if this is going to work, but how it works is that the `playwright-teamcity-reporter` sends test results directly to TeamCity, allows us to see the test reports directly within the TeamCity interface.

### Checklist

### DIFF
--- a/frontend/playwright.config.ts
+++ b/frontend/playwright.config.ts
@@ -10,7 +10,7 @@ export default defineConfig({
   forbidOnly: !!process.env.CI,
   retries: process.env.CI ? 2 : 0,
   workers: process.env.CI ? 1 : undefined,
-  reporter: 'html',
+  reporter: [['playwright-teamcity-reporter']],
   use: { baseURL: `http://localhost:${port}/`, trace: 'on-first-retry' },
   projects: [{ name: 'chromium', use: { ...devices['Desktop Chrome'] } }],
   webServer: {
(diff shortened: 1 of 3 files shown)


frontend/playwright.config.ts
```diff
@@ -10,7 +10,7 @@ export default defineConfig({
   forbidOnly: !!process.env.CI,
   retries: process.env.CI ? 2 : 0,
   workers: process.env.CI ? 1 : undefined,
-  reporter: 'html',
+  reporter: [['playwright-teamcity-reporter']],
```
gregory-j-baker, line 13: Would it make sense to keep `html`?
ex:
```
reporter: ['html', 'playwright-teamcity-reporter'],
```
(or whatever the correct syntax is)

gregory-j-baker (approved):

frontend/playwright.config.ts
sebastien-comeau: Maybe we  should use `process.env.CI` to use different reporters.

```ts
const reporter: ReporterDescription[] = [['list'], ['html']]

export default defineConfig({
  ...
  reporter: process.env.CI ? [...reporter, ['playwright-teamcity-reporter']]: reporter,
  ...
```

sebastien-comeau (approved):

gregory-j-baker (approved):

PR #2083, opened 2024-08-22
disable parent-guardian radio on edit

Fbasham:
### Description
take advantage of state variable `isNew` to make the parent/legal guardian question in the child information screens readonly if and only if the child isn't newly created.

[AB#4401](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/4401)


sebastien-comeau (approved):
👍

PR #2106, opened 2024-08-28
4453 persist child date of birth only if it is valid

Fbasham:
### Description
The child's age should be persisted if and only if it's valid (i.e. their age category is either `'youth'` or `'children'`.

AB#4453

### Test Instructions
- Perform the happy flow (staging) for my children up to children's application page.
- Click Add a child, and enter first name, last name, DOB (01/01/2020), no SIN, select Yes (parent or legal guardian), and click Continue 
- Select no (dental insurance or coverage), and click Continue 
- Select no (federal, provincial, territorial benefits), and click Continue
- Edit child's information, and update the DOB year to 1950, and click Continue (then Back twice (on cannot-apply-child page))

Expected Result: The year value of 1950 shouldn't be stored in the session (2020 should be used instead) as 1950 isn't valid.


frontend/app/routes/$lang/_public/apply/$id/adult-child/children/$childId/information.tsx
```diff
@@ -182,13 +182,19 @@ export async function action({ context: { session }, params, request }: ActionFu
     });
   }
 
+  const ageCategory = getAgeCategoryFromDateString(parsedDataResult.data.dateOfBirth);
```
nicholas-ly, line 185: Kind of a nitpick but this identical function call to `getAgeCategoryFromDateString(..)` occurs below on line 210. Consider refactoring to have a single function call and reusing the `ageCategory` local variable throughout.

This same comment can be applied to the `child` flow's `information.tsx`.

Fbasham: good catch

nicholas-ly (approved):

PR #2163, opened 2024-09-18
Set up OWASP dependency check

shawn320:
### Description

Added a Maven profile in pom.xml, this profile will use the `dependency-check-maven` plugin to scan the dependencies during build

### Checklist

### DIFF
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -300,17 +300,9 @@
 				</configuration>
 			</plugin>
 			<plugin>
-				<!-- Perform CVE dependency scans during the verify phase -->
 				<groupId>org.owasp</groupId>
 				<artifactId>dependency-check-maven</artifactId>
 				<version>${dependency-check-maven.version}</version>
-				<executions>
-					<execution>
-						<goals>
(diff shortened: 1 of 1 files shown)


gregory-j-baker (commented):
I wouldn't add a maven profile for this.. instead, just configure plugin but don't bind it to any lifecycle phases.

It can then be executed on demand with `mvn dependency-check-maven:check` (or something like that).

backend/pom.xml
```diff
@@ -300,5 +300,27 @@
...
+	<profiles>
+		<profile>
+			<id>dependency-check</id>
+			<build>
+				<plugins>
+					<plugin>
+						<!-- Perform CVE dependency scans during the verify phase -->
+						<groupId>org.owasp</groupId>
+						<artifactId>dependency-check-maven</artifactId>
+						<version>${dependency-check-maven.version}</version>
```
nicholas-ly, line 312: A new field within `<properties>` for `dependency-check-maven.version` should be introduced (the latest version is 10.0.4).

nicholas-ly (dismissed):

PR #2210, opened 2024-09-26
refactor(frontend): rename preferred communication service methods

Fbasham:
### Description
per title

[AB#4573](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/4573)

### DIFF
--- a/frontend/app/routes/public/apply/$id/child/review-adult-information.tsx
+++ b/frontend/app/routes/public/apply/$id/child/review-adult-information.tsx
@@ -74,8 +74,7 @@ export async function loader({ context: { configProvider, serviceProvider, sessi
   const countryHome = serviceProvider.getCountryService().getCountryById(state.contactInformation.homeCountry);
 
   // Getting CommunicationPreference by Id
-  const communicationPreference = serviceProvider.getPreferredCommunicationMethodService().findById(state.communicationPreferences.preferredMethod);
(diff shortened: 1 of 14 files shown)


frontend/app/routes/public/apply/$id/child/review-adult-information.tsx
```diff
@@ -74,7 +74,7 @@ export async function loader({ context: { configProvider, serviceProvider, sessi
   const countryHome = serviceProvider.getCountryService().getCountryById(state.contactInformation.homeCountry);
 
   // Getting CommunicationPreference by Id
-  const communicationPreference = serviceProvider.getPreferredCommunicationMethodService().findById(state.communicationPreferences.preferredMethod);
+  const communicationPreference = serviceProvider.getPreferredCommunicationMethodService().findPreferredCommunicationMethodsById(state.communicationPreferences.preferredMethod);
   invariant(communicationPreference, `Unexpected communication preference: ${state.communicationPreferences.preferredMethod}`);
```
nicholas-ly, line 78: **Suggestion:** Having the service expose a `get(..)` instead of a `find(..)`, based on our conventions, would mean that the service would throw if no `communicationPreference` was found rather than returning `null`.

Throwing early in the service would eliminate the need for all the `invariant` checks afterwards where `findPreferredCommunicationMethodsById(..)` is currently called.

nicholas-ly (approved):

PR #2309, opened 2024-10-17
status checker back button fix

nibivi77:
### Description
Thanks for the help @sebastien-comeau 

[AB#4729](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/4729)


sebastien-comeau (approved):
Take note that we have to update the `getPathById` once `v3.4.0` is merged into main because the route id is different.

PR #2357, opened 2024-10-24
Adding contact info ITA french translations.

nibivi77:
[AB#4743](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/4743)

### DIFF
--- a/frontend/public/locales/fr/renew-ita.json
+++ b/frontend/public/locales/fr/renew-ita.json
@@ -2,11 +2,11 @@
   "contact-information": {
     "page-title": "Coordonnées",
     "phone-header": "Numéro de téléphone",
-    "add-phone": "(FR) Adding a phone number helps Government of Canada contact you if there is an issue with your renewal. Otherwise, we'll contact you by mail.",
+    "add-phone": "L'ajout d'un numéro de téléphone à jour au dossier aidera le gouvernement du Canada à communiquer avec vous en cas de problème lié à votre demande de renouvellement. Sinon, nous communiquerons avec vous par la poste.",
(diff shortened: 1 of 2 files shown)


frontend/public/locales/fr/renew-ita.json
```diff
@@ -20,9 +20,7 @@
       "confirm-email-required": "Confirmez l'adresse courriel",
       "email-required": "Entrez l'adresse courriel",
       "email-valid": "Entrez l'adresse courriel dans le bon format, par exemple nom@exemple.com",
-      "confirm-email-valid": "Entrez la confirmation de l'adresse courriel dans le bon format, par exemple nom@exemple.com",
-      "preferred-language-required": "Sélectionnez la langue de communication préférée",
-      "charactrs-valid": "Seuls les lettres, les chiffres et les caractères suivants sont acceptés\u00a0: apostrophe ('), virgule (,), point (.), trait d'union (-) et parenthèse ()"
+      "confirm-email-valid": "Entrez la confirmation de l'adresse courriel dans le bon format, par exemple nom@exemple.com"
```
gregory-j-baker, line 23: Bit of a nitpick, but I wonder if this should be `nom@example.com`. There's a reason we use `example.com` everywhere: https://en.wikipedia.org/wiki/Example.com

`exemple.com` is currently being squatted, but in the future it could be a malicious site.. we don't want to reference a potentially malicious site in the application.

nibivi77: Fair point!

gregory-j-baker (approved):

PR #2361, opened 2024-10-25
fix duplicate headings and back navigations in renew-adult-child flow

Fbasham:
### Description
per title

[AB#4586](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/4586)

### DIFF
--- a/frontend/app/routes/public/renew/$id/adult-child/confirm-email.tsx
+++ b/frontend/app/routes/public/renew/$id/adult-child/confirm-email.tsx
@@ -169,8 +169,7 @@ export default function RenewAdultChildConfirmEmail() {
         <errorSummary.ErrorSummary />
         <fetcher.Form method="post" noValidate>
           <input type="hidden" name="_csrf" value={csrfToken} />
-          <fieldset className="mb-6">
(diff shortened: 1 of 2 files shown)


frontend/app/routes/public/renew/$id/adult-child/confirm-email.tsx
```diff
@@ -170,7 +170,6 @@ export default function RenewAdultChildConfirmEmail() {
         <fetcher.Form method="post" noValidate>
           <input type="hidden" name="_csrf" value={csrfToken} />
           <fieldset className="mb-6">
-            <legend className="mb-4 font-lato text-2xl font-bold">{t('renew-adult-child:confirm-email.email-header')}</legend>
```
nicholas-ly, line 173: I believe we can remove this parent `<fieldset>` now as there's no corresponding `<legend>` for it.

(same for `/confirm-phone`).

Fbasham: true.  We still need margin and it doesn't appear that the Input component accepts additional styles as props, so I'll just keep the div in place of fieldset.

PR #2389, opened 2024-10-29
add demographic-survey/summary

Fbasham:
### Description
adds:
- `demographic-survey/summary`
- `demographic-survey/questions/$memberId`
- empty `demographic-survey/submitted`

NOTES:
- is it possible to have `demographic-survey/questions?memberId=${memberId}` (i.e. using search params?), I tried, but we get an error on application startup since it appears the routes need to be statically generated beforehand?\
- dummy data is used for the sake of displaying something. This means we'll need to create a new stateful object, like `DemographicSurveyState`, which will be initialized with names, ids, etc, on the `/confirmation` route in action when a user clicks the `Take Survey` button in the contextual alert.  We'll also likely need a route-helper for this app so we can perform operations on the state

AB#4571

image

image


gregory-j-baker (approved):

sebastien-comeau:
> is it possible to have demographic-survey/questions?memberId=${memberId} (i.e. using search params?), I tried, but we get an error on application startup since it appears the routes need to be statically generated beforehand?

```ts
const url = new URL(getPathById('demographic-survey/questions'));
url.searchParams.set("memberId", memberId);
const resolvedUrl = url.toString();
```

gregory-j-baker:
> > is it possible to have demographic-survey/questions?memberId=${memberId} (i.e. using search params?), I tried, but we get an error on application startup since it appears the routes need to be statically generated beforehand?
> 
> ```ts
> const url = new URL(getPathById('demographic-survey/questions'));
> url.searchParams.set("memberId", memberId);
> const resolvedUrl = url.toString();
> ```

Query params are not a part of the route path. In the example you gave, the path would be `/:lang/demographic-survey/questions` and the query string would be extracted by the loader.

See https://github.com/DTS-STN/canadian-dental-care-plan/blob/main/frontend/app/routes/api/health.ts#L32 for an example.

Fbasham:
> > is it possible to have demographic-survey/questions?memberId=${memberId} (i.e. using search params?), I tried, but we get an error on application startup since it appears the routes need to be statically generated beforehand?
> 
> ```ts
> const url = new URL(getPathById('demographic-survey/questions'));
> url.searchParams.set("memberId", memberId);
> const resolvedUrl = url.toString();
> ```

@sebastien-comeau ,
I'd be looking to do this for an `InlineLink`.  Is it still possible?

https://github.com/DTS-STN/canadian-dental-care-plan/pull/2389/files#diff-d3cb8064fd4341f432a9dd0823c891db1f6d9c79d88ab8ab0be2fb287a7e9ff6R81

sebastien-comeau:
The demographic survey should have his own `stateId` similar to the online application.

E.g.: `demographic-survey/{stated-id}/questions/`

As for the members we can also to the same as the online application children if needed or keep it as is with the search params.

E.g.: `demographic-survey/{stated-id}/members/{member-id}/questions/`

Don't forget the Adobe Analytics team will ask us to transform the url sent to them.

sebastien-comeau:
@Fbasham, @gregory-j-baker  is right. In you code you're using `params` instead of `searchParams`.

The current code is passing `memberId` as param to InlineLink and should work.

```tsx
<InlineLink routeId="public/demographic-survey/questions/$memberId" params={{ ...params, memberId: member.id.toString() }}>
```

PR #2438, opened 2024-11-01
[gitops] enable trace in uat1

gregory-j-baker:
Title.


gregory-j-baker:
Note: bypassing review since this is trivial.

PR #2472, opened 2024-11-06
Adding link to the image header in public route.

nibivi77:


nicholas-ly (approved):

frontend/app/components/layouts/public-layout.tsx
sebastien-comeau, line 1: Was it removed? The logo as a link in PROD.

nicholas-ly: I asked the same question.

The logo should be a link in the public layout, which this PR adds back.
The link has been removed from the logo in the protected layout (see `page-header-brand` changes): #2468

sebastien-comeau (approved):

PR #2479, opened 2024-11-06
change client-number format

Fbasham:
### Description
per Guillaume's request in slack


sebastien-comeau (commented):
Is the string returned to the server contains spaces?

PR #2498, opened 2024-11-07
Adding conditional link to header logo

nibivi77:
### Description
In `public`, the logo should redirect users to canada.ca
In `protected`, the logo should not have a link

AB#4788


frontend/app/components/page-header-brand.tsx
```diff
@@ -1,20 +1,37 @@
...
+  const pageHeaderLogo = <img className="h-8 w-auto" src={`/assets/sig-blk-${i18n.language}.svg`} alt={t('gcweb:header.govt-of-canada.text')} property="logo" width="300" height="28" decoding="async" />;
+  const readOnlyLogo = <span className="sr-only">/ {altLogoContent}</span>;
   return (
     <div id="wb-bnr">
       <div className="container flex items-center justify-between gap-6 py-2.5 sm:py-3.5">
         <div property="publisher" typeof="GovernmentOrganization">
-          <img className="h-8 w-auto" src={`/assets/sig-blk-${i18n.language}.svg`} alt={t('gcweb:header.govt-of-canada.text')} property="logo" width="300" height="28" decoding="async" />
-          <span className="sr-only">/ {altLogoContent}</span>
+          {link ? (
+            <Link to="https://canada.ca/" property="url">
```
sebastien-comeau, line 24: Use normal `a` html tag.

```suggestion
            <a href="https://canada.ca/" property="url">
```

frontend/app/components/page-header-brand.tsx
```diff
@@ -1,20 +1,37 @@
...
+}
+
+export function PageHeaderBrand({ link = false }: PageHeaderBrandProps) {
   const { i18n, t } = useTranslation(['gcweb']);
 
   const altLanguage = getAltLanguage(i18n.language);
   const altLogoContent = <span lang={altLanguage}>{i18n.getFixedT(altLanguage)('gcweb:header.govt-of-canada.text')}</span>;
-
+  const pageHeaderLogo = <img className="h-8 w-auto" src={`/assets/sig-blk-${i18n.language}.svg`} alt={t('gcweb:header.govt-of-canada.text')} property="logo" width="300" height="28" decoding="async" />;
+  const readOnlyLogo = <span className="sr-only">/ {altLogoContent}</span>;
```
sebastien-comeau, line 18: Use one variable since they are always grouped together.

```suggestion
  const headerLogo = (<>
    <img className="h-8 w-auto" src={`/assets/sig-blk-${i18n.language}.svg`} alt={t('gcweb:header.govt-of-canada.text')} property="logo" width="300" height="28" decoding="async" />;
    <span className="sr-only">/ {altLogoContent}</span>;
  </>);
```

sebastien-comeau (changes requested):

gregory-j-baker (approved):
I think that the url should be configurable, and not hardcoded.. we can fix this in a future PR (or not).

sebastien-comeau:
> I think that the url should be configurable, and not hardcoded.. we can fix this in a future PR (or not).

Also, should link be localized based on URL route context?

frontend/app/components/page-header-brand.tsx
```diff
@@ -3,18 +3,33 @@ import { useTranslation } from 'react-i18next';
...
       <div className="container flex items-center justify-between gap-6 py-2.5 sm:py-3.5">
         <div property="publisher" typeof="GovernmentOrganization">
-          <img className="h-8 w-auto" src={`/assets/sig-blk-${i18n.language}.svg`} alt={t('gcweb:header.govt-of-canada.text')} property="logo" width="300" height="28" decoding="async" />
-          <span className="sr-only">/ {altLogoContent}</span>
+          {link ? (
+            <a href="https://canada.ca/" property="url">
+              {headerLogo}
+            </a>
+          ) : (
+            <>{headerLogo}</>
```
sebastien-comeau, line 30: React Fragment in unecessary.

```suggestion
            headerLogo
```

sebastien-comeau (changes requested):

frontend/app/components/page-header-brand.tsx
```diff
@@ -3,18 +3,33 @@ import { useTranslation } from 'react-i18next';
 import { LanguageSwitcher } from '~/components/language-switcher';
 import { getAltLanguage } from '~/utils/locale-utils';
 
-export function PageHeaderBrand() {
+export interface PageHeaderBrandProps {
+  link?: boolean;
```
sebastien-comeau, line 7: I don't think 'link' is the best choice here. I would recommend using a more explicit name for the prop and making it required.

My suggestion is to use a prop called headerLogoVariant with two options.

```suggestion
  /**
   * Determines how the logo is displayed in the header.
   * 
   * @type {'logo-only' | 'logo-with-link'}
   * @description
   *   - `'logo-only'`: Displays the logo without a link.
   *   - `'logo-with-link'`: Displays the logo wrapped in a link.
   */
  headerLogoVariant: 'logo-only' | 'logo-with-link';
```

OR

`headerLogoUrl` prop: If this prop is defined and non-empty,  the logo will be wrapped in a link pointing 
to this URL.

```suggestion
  /**
   * The URL to wrap the logo with. If this prop is defined and non-empty, 
   * the logo will be wrapped in a link pointing to this URL.
   * 
   * @type {string}
   */
  headerLogoUrl?: string;
```

sebastien-comeau (changes requested):

frontend/app/components/page-header-brand.tsx
gregory-j-baker: imo the `headerLogo` prefix is kinda redundant.. just use `variant`
alternatively, why not move the link to the parent component? `<Link><PageHeaderBrand/></Link>`

nibivi77: I don't think we want to wrap the whole `<PageHeaderBrand>` We just want to wrap the logo itself. So I'll take the `variant` prop approach.

frontend/.env.example
```diff
@@ -211,6 +211,14 @@ ECAS_BASE_URI=https://srv136.services.gc.ca/ecas-seca/rascl/SCL
 # (optional; default https://service.canada.ca)
 SCCH_BASE_URI=https://secure-client-hub-dev.bdm.dshp-phdn.net
 
+# English Header Logo URL -- used as a link when clicking header logo
+# (optional; default https://canada.ca/en)
+SCCH_BASE_URI_EN=https://canada.ca/en
```
sebastien-comeau, line 216:
```suggestion
HEADER_LOGO_URL_EN=https://canada.ca/en
```

frontend/.env.example
```diff
@@ -211,6 +211,14 @@ ECAS_BASE_URI=https://srv136.services.gc.ca/ecas-seca/rascl/SCL
...
+
+# French Header Logo URL -- used as a link when clicking header logo
+# (optional; default https://canada.ca/fr)
+SCCH_BASE_URI_FR=https://canada.ca/fr
```
sebastien-comeau, line 220:
```suggestion
HEADER_LOGO_URL_FR=https://canada.ca/fr
```

sebastien-comeau (changes requested):

sebastien-comeau (approved):

PR #2505, opened 2024-11-08
4810 only display valid members in demographic survey

Fbasham:
### Description
When a user clicks "Renew a child" in `renew/$id/adult-child/children` (the children summary hub), a new entry in state is created for the child.  The user is then brought to the information screen to enter the child's info.  The user can however click back, which keeps this entry in state, and as a consequence, no information for the child has been entered, meaning they're invalid.

There are few ways to handle only displaying valid members in the demographic survey.  I've landed on this approach since it appears to be the simplest (all members need a first and last name)

[AB#4810](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/4810)

### DIFF
--- a/frontend/app/route-helpers/demographic-survey-route-helpers.server.ts
+++ b/frontend/app/route-helpers/demographic-survey-route-helpers.server.ts
@@ -6,6 +6,7 @@ import { randomUUID } from 'crypto';
 import { z } from 'zod';
 
 import type { RenewState } from './renew-route-helpers.server';
+import { getChildrenState } from './renew-route-helpers.server';
 import { getLocaleFromParams } from '~/utils/locale-utils.server';
 import { getLogger } from '~/utils/logging.server';
 import { getPathById } from '~/utils/route-utils';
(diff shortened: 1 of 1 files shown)


sebastien-comeau (commented):
I suggest to create a reusable function similar to the online application children `isNewChildState`. Something like `isChildStateCompleted(state: ChildrenState): boolean`. You can also remove uncompleted children states on specific actions, maybe when review page is loading.

Fbasham:
> I suggest to create a reusable function similar to the online application children `isNewChildState`. Something like `isChildStateCompleted(state: ChildrenState): boolean`. You can also remove uncompleted children states on specific actions, maybe when review page is loading.

okay, why not just reuse the `getChildren` function.  Functions are intended to be reuseable after all.

sebastien-comeau (approved):

sebastien-comeau:
> > I suggest to create a reusable function similar to the online application children `isNewChildState`. Something like `isChildStateCompleted(state: ChildrenState): boolean`. You can also remove uncompleted children states on specific actions, maybe when review page is loading.
> 
> okay, why not just reuse the `getChildren` function. Functions are intended to be reuseable after all.

Good observation. I wasn't aware that existed in the renewal flow. 👍

PR #2586, opened 2024-11-18
add children demographic survey route to protected renew application

Fbasham:
### Description
adds `protected/renew/$id/$childId/demographic-survey`

I reused the locales from `protected/renew/$id/demographic-survey` since they're identical and in the same file.  Let me know if you'd like a separate entry for this route.

[AB#4822](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/4822)


gregory-j-baker (approved):
I think it might be smart to separate the localized strings (instead of reusing existing ones) just in case the two flows diverge in the future.

However, I think that can be done later on.

PR #2653, opened 2024-11-29
4862 use findClientApplicationBySin to seed initial protected renew state

Fbasham:
### Description
- calls `findClientApplicationBySin` and populated the applicant information and children information in state

I believe we might need to create a mapper function which will map the immutable client application from the service method call to the protected renew state since they differ in shape and we need more than just the fields initially pulled in this PR for the review/edit screens.

AB#4682

image

### Checklist


frontend/app/.server/routes/helpers/protected-renew-route-helpers.ts
```diff
@@ -196,24 +200,31 @@ export function clearProtectedRenewState({ params, session }: ClearStateArgs) {
...
  */
-export function startProtectedRenewState({ id, session }: StartArgs) {
+export async function startProtectedRenewState({ id, session, appContainer }: StartArgs) {
   const log = getLogger('protected-renew-route-helpers.server/startProtectedRenewState');
   const parsedId = idSchema.parse(id);
+  const sessionName = getSessionName(parsedId);
+
+  const clientApplicationService = appContainer.get(TYPES.domain.services.ClientApplicationService);
+  const clientApplication = await clientApplicationService.findClientApplicationBySin(session.get('userInfoToken').sin);
+  invariant(clientApplication, 'Expected client application to be defined');
```
nicholas-ly, line 218: We can redirect to the `/data-unavailable` route if no client application is found.

frontend/app/.server/routes/helpers/protected-renew-route-helpers.ts
```diff
@@ -196,24 +200,31 @@ export function clearProtectedRenewState({ params, session }: ClearStateArgs) {
...
 
+  // TODO: create a mapper function from clientApplication to initialState?
   const initialState: ProtectedRenewState = {
     id: parsedId,
     editMode: false,
-    children: [],
+    applicantInformation: clientApplication.applicantInformation,
+    children: clientApplication.children.map((child) => ({ id: randomUUID(), ...child, ...child.information })),
   };
```
nicholas-ly, line 226: My thinking is that we store the immutable `clientApplication` as its own object in state and any further modifications will be kept as a separate object(s). 

The state mapper will take the unmodified `clientApplication` and any new data to build the new renewal DTO, similar to how it's done in the unauthenticated area.

The member selection can reference this `clientApplication` object in state, whereas the review information can originally reference this object and render the new data in state depending on if the user updates their info.

Fbasham: hmm, okay.  I'll have to go back to the drawing board.  I understand you though.

Fbasham: alright, so I tried some different permutations, but came to the same conclusion.  We have fields on both `applicantInformation` and `children` (`externallyReviewed`, `previouslyReviewed`, `firstName`, `lastName`, `id` (for the child that's part of the path param)) that are coupled to the member-selection screen. Before we land on this screen, we need values populated in state.  Ultimately I think we're going to need a mapper, but that's going to be a pretty big effort for this PR.  My intention is to for this mapper to be called in the `startProtectedRenewState` as in this PR.  I'm not sure where `externallyReviewed`, `previouslyReviewed` are coming from, or if those are native to our application/state.  Either way, this is more complicated than it initially appears.

Fbasham: For the primary applicant that's doable, but for the children I don't think there's an easy work around, unless I'm missing something obvious.

nicholas-ly (approved):

nicholas-ly (approved):

PR #2668, opened 2024-12-01
fix: broken terraform version

gregory-j-baker:
### Description

Terraform v1.10.x adds ephemeral variables, which breaks Terragrunt. This PR pins the version to a maximum of v1.9.x until a fix is released.


gregory-j-baker:
Closing this since there should be a fix incoming [here](/hashicorp/terraform/pull/36137)

PR #2713, opened 2024-12-09
add confirmation route to protected renew application

Fbasham:
### Description
- adds the confirmation screen to the protected renew app

[AB#4822](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/4822)

### Checklist

### DIFF
--- a/frontend/app/.server/routes/helpers/protected-renew-route-helpers.ts
+++ b/frontend/app/.server/routes/helpers/protected-renew-route-helpers.ts
@@ -256,12 +256,6 @@ export async function startProtectedRenewState({ id, session, appContainer }: St
         id: randomUUID(),
         information: child.information,
         dentalInsurance: immutableChild?.dentalInsurance,
-        dentalBenefits: {
-          hasFederalBenefits: Boolean(immutableChild?.dentalBenefits[0]),
-          federalSocialProgram: immutableChild?.dentalBenefits[0],
-          hasProvincialTerritorialBenefits: Boolean(immutableChild?.dentalBenefits[1]),
-          provincialTerritorialSocialProgram: immutableChild?.dentalBenefits[1],
-        },
       };
       return childStateObj;
     }),
--- a/frontend/app/page-ids.ts
+++ b/frontend/app/page-ids.ts
@@ -24,6 +24,7 @@ export const pageIds = {
       reviewAdultInformation: 'CDCP-PROT-RENW-0017',
       reviewChildInformation: 'CDCP-RENW-ADCH-0018',
       reviewSubmit: 'CDCP-PROT-RENW-0019',
+      confirmation: 'CDCP-PROT-RENW-0020',
     },
     dataUnavailable: 'CDCP-PROT-0099',
   },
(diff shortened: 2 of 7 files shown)


frontend/app/.server/resources/power-platform/client-application.json
```diff
@@ -106,10 +106,10 @@
           {
             "InsurancePlanIdentification": [
               {
-                "IdentificationID": "b5f25fea-a7a9-ee11-a569-000d3af4f898"
+                "IdentificationID": "e174250d-26c5-ee11-9079-000d3a09d640"
               },
               {
-                "IdentificationID": "1788f1db-25c5-ee11-9079-000d3a09d640"
+                "IdentificationID": "b3f25fea-a7a9-ee11-a569-000d3af4f898"
```
nicholas-ly, line 112: I think this change is a result the service throwing because of the following code: https://github.com/DTS-STN/canadian-dental-care-plan/blob/6bde044cd8cb9b50ab3b9185827625a2dd4b0965/frontend/app/routes/protected/renew/%24id/review-adult-information.tsx#L135

The structure of `clientApplication.dentalBenefits` (the client's info on file), is different from the structure of `dentalBenefits` in the protected state. `clientApplication.dentalBenefits` is an array of IDs with no way to distinguish whether the benefit is a federal or provincial/territorial benefit unless we call both their respective services. Since the first item is a provincial benefit ID, getting a federal benefit by that ID will throw.

I think we should add a `TODO` on that line in `review-adult-information`.

- We should introduce `findBy..` methods in their respective services so that the services won't throw but return `null` instead if a benefit cannot be found by an ID.
- Or we can combine the federal and provincial/territorial benefits service into a single dental insurance service - the JSON file in SharePoint actually has all of these benefits (federal, provincial, and territorial) rolled up into a single JSON response.

Fbasham: It's because of the `review-child-information`.  My brain is barely functioning right now FWIW, but I'm not sure where the best place to solve this is.  I wonder if it's guaranteed that the `clientApplication` always returns an array of length 2, and if the index of the array indicates whether it's provincial or federal.  I think we still need to map the `clientApplication` to initial state, especially for the child state.  I did make a todo for that.  For now I need to think of a better way to handle the child benefits coming from the `clientApplication`.  The changes I made do work though.

nicholas-ly: `clientApplication` can return 0, 1 or 2 elements. Currently, there's no way to guarantee that the first element is federal or provincial/territorial.

Regarding mapping `clientApplication` to initial state, I made this comment for `review-adult-information`:
https://github.com/DTS-STN/canadian-dental-care-plan/pull/2706#discussion_r1873300796. This change won't impact the state object, only the object the `loader(..)` returns. We should be able to do the same for `review-child-information`.

My understanding is that the `review-child-information` page should display everything on file in `clientApplication` initially, except the private dental insurance and demographics.

Then if the user modifies anything, the application makes the change and saves it in the corresponding state object.

nicholas-ly (approved):

nicholas-ly (approved):

PR #2735, opened 2024-12-11
fix address state changes in protected renew app

Fbasham:
### Description
- before this PR no changes to the `homeAddress` and `mailingAddress` were being made in the edit screen.  
- we could at some point remove `addressInformation` from the `ProtectedRenewState` in a future PR?
- should the edit screens fallback to the `clientApplication` data for addresses?  

AB#4822
### Checklist


nicholas-ly (approved):
I think falling prepopulating with current information is good UX, which reminds me of some changes we made yesterday with partner application. 

We should revisit that form and prepopulate the partner info, but also ensure that if the applicant selects a marital status that doesn't require partner info, that data is cleared.

PR #2785, opened 2024-12-17
AB#4587 -- Added renew child review parent information page

AdamP-HRSDC-RHDCC:
### Description
Added renew child review parent information page

[AB#4587](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/4587)

rpi


frontend/app/.server/routes/helpers/renew-child-route-helpers.ts
```diff
@@ -171,16 +174,20 @@ export function validateRenewChildStateForReview({ params, state }: ValidateStat
...
   }
 
-  if (hasAddressChanged && addressInformation === undefined) {
-    throw redirect(getPathById('public/renew/$id/child/update-address', params));
+  if (hasAddressChanged && isHomeAddressSameAsMailingAddress && mailingAddress === undefined) {
+    throw redirect(getPathById('public/renew/$id/child/update-mailing-address', params));
+  }
+
+  if (hasAddressChanged && !isHomeAddressSameAsMailingAddress && homeAddress === undefined) {
+    throw redirect(getPathById('public/renew/$id/child/update-mailing-address', params));
```
nibivi77, line 182: I might be wrong here, but what if we make the two `if` statements into this:

`if (hasAddressChanged && isHomeAddressSameAsMailingAddress && mailingAddress === undefined){}`

nicholas-ly: I agree with your comment that we can just do `if (hasAddressChanged && mailingAddress === undefined)`, since if the applicant's address changed, they must enter a mailing address. 

I think that the second condition should redirect to `/update-home-address` instead.

Fbasham (commented):
for the sake of continuity, the file and route should be `review-adult-information`

frontend/app/.server/routes/helpers/renew-child-route-helpers.ts
AdamP-HRSDC-RHDCC: Removed the "isHomeAddressSameAsMailingAddress" from the first conditional and fixed the redirect on the second one to target the home address page.

AdamP-HRSDC-RHDCC:
> for the sake of continuity, the file and route should be `review-adult-information`

I've updated all parent references to adult, previous reasoning behind this was to match with page descriptions/titles, but I agree keeping in line with the existing adult naming convention would work better.

Fbasham (approved):

PR #2820, opened 2024-12-20
fix state issues in protected-renew app

Fbasham:
### Description
Proposed fixes for some 'bugs' in the protected space

Most of these issues happen because of the nullish coalesce with the clientApplication data.  I feel like we need to initially seed the state from the clientApplication in the `startProtectedRenewState` function in the route-helper, especially for mutable data.  

Note: the phone numbers aren't initially populated in their respective fields because the mock data we have uses fake data that errors based on the phone validation library we use.  I also don't think the "no update" text in the review screen is necessary, based on Figma

[AB#5027](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/5027), [AB#5026](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/5026), [AB#5028](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/5028)

### Checklist

### DIFF
--- a/frontend/app/.server/routes/helpers/protected-renew-route-helpers.ts
+++ b/frontend/app/.server/routes/helpers/protected-renew-route-helpers.ts
@@ -249,6 +249,12 @@ export function startProtectedRenewState({ applicationYear, clientApplication, i
     editMode: false,
     applicationYear,
     clientApplication,
+    contactInformation: {
+      phoneNumber: clientApplication.contactInformation.phoneNumber,
+      phoneNumberAlt: clientApplication.contactInformation.phoneNumberAlt,
+      email: clientApplication.contactInformation.email,
+    },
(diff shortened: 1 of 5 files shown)


nicholas-ly (approved):
> Most of these issues happen because of the nullish coalesce with the clientApplication data. I feel like we need to initially seed the state from the clientApplication in the startProtectedRenewState function in the route-helper, especially for mutable data.

The more I looked into this flow to deal with similar issues, the more suggestion make sense. My bad, I may or may not have removed a previous `TODO` in the route helper that suggests this.

PR #2928, opened 2025-01-10
Refactoring AddressValidation to a single generic service.

nibivi77:
### Description
Refactoring AddressValidation to a single generic service.

[AB#5032](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/5032)

### Checklist

### Test Instructions
Try `update-mailing-address` in all 4 public routes (ita/adult/adult-child/child)


frontend/app/routes/public/routes.ts
```diff
@@ -4,16 +4,6 @@ export const routes = [
...
-      {
-        id: 'public/address-validation/index',
-        file: 'routes/public/address-validation/index.tsx',
-        paths: { en: '/:lang/address-validation', fr: '/:lang/validation-adresse' },
-      },
-      {
-        id: 'public/address-validation/review',
-        file: 'routes/public/address-validation/review.tsx',
-        paths: { en: '/:lang/address-validation/review', fr: '/:lang/validation-adresse/revue' },
-      },
```
nicholas-ly, line 16: I don't know if we should be removing this `address-validation` PoC routes just yet. @sebastien-comeau recently used it to add route module type safety in #2912, which we can use as a guide for other routes.

frontend/app/.server/routes/public/address-validation/mailing-address.validator.factory.ts
```diff
@@ -16,7 +17,7 @@ export interface MailingAddressValidatorFactory {
    * @param locale The locale to use for validation.
    * @returns A new mailing address validator.
    */
-  createMailingAddressValidator(locale: AppLocale): MailingAddressValidator;
+  createMailingAddressValidator(errorMessages: AddressValidatorErrorMessages): MailingAddressValidator;
```
nicholas-ly, line 20: Might be a nitpick but we should update this interface method documentation as it still references `locale` rather than `errorMessages`. The same comment applies for the `createMailingAddressValidator(..)` implementation down below.

nibivi77: Good catch!

sebastien-comeau: This is a factory to create a new AddressValidations's MailingAddressValidator. It requires you to pass the `locale` and it take care to initialize the `MailingAddressValidator` with the AddressValidations's MailingAddressValidator error messages from i18n.

nibivi77: What do you suggest instead?

sebastien-comeau: Move mailing address error messages in a `common` namespace if they are all the same and continue to use the factory has it was before (passing the current request `locale`). If they have differences then what you did is good. Uou can keep the factory method with `locale` that will use the `common` error messages and the factory method that with the error messages if for some cases they are different.

Another thing: Move `mailing-address.validator.factory.ts` and `mailing-address.validator.factory.ts` to `app/.server/routes/validators` since they are now common routes validator.

nibivi77: Sounds good! On it. Thanks Seb!

PR #2949, opened 2025-01-14
fix navigation and payload with single child renew behind msca

Fbasham:
### Description
If the user submits renewal ONLY for their dependent, then the `review-child-information` screen should be displayed before the `review-adult-information` screen.  The dental benefits section on the `review-adult-information` screen should not be displayed and the payload needs to be modified to remove the dental benefits for the primary applicant.  

[AB#5087](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/5087)

### Checklist

### DIFF
--- a/frontend/app/.server/routes/helpers/protected-renew-route-helpers.ts
+++ b/frontend/app/.server/routes/helpers/protected-renew-route-helpers.ts
@@ -369,11 +369,11 @@ export function loadProtectedRenewStateForReview({ params, request, session, dem
 }
 
 export function isPrimaryApplicantStateComplete(state: ProtectedRenewState, demographicSurveyEnabled: boolean) {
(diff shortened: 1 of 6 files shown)


nicholas-ly (approved):
Navigation looks good to me. One issue I found that can be addressed in a future PR:
- Complete the renewal for a child until you reach `/review-child-information`
- Hit "Back" and complete the renewal for the adult until you reach `/review-adult-information`
- Notice that in the payload, `PrivateDentalInsuranceIndicator` does not exist and `InsurancePlanIdentification` is empty - these values should be set (the former being the form input by the adult and the latter being what's on file in `clientApplication`) because we are completing renewal for both adult and children

PR #3047, opened 2025-01-22
Fix phone number save state

AdamP-HRSDC-RHDCC:
### Description
Fixed phone number not saving to state across all flows

parsedResults does not contain the phone number fields if they are not explicitly set in the phoneNumberSchema transform segment

image

### DIFF
--- a/frontend/app/routes/public/apply/$id/adult-child/contact-information.tsx
+++ b/frontend/app/routes/public/apply/$id/adult-child/contact-information.tsx
@@ -177,6 +177,8 @@ export async function action({ context: { appContainer, session }, params, reque
       ...val,
       homePostalCode: val.homeCountry && val.homePostalCode ? formatPostalCode(val.homeCountry, val.homePostalCode) : val.homePostalCode,
       mailingPostalCode: val.mailingCountry && val.mailingPostalCode ? formatPostalCode(val.mailingCountry, val.mailingPostalCode) : val.mailingPostalCode,
+      phoneNumber: val.phoneNumber,
+      phoneNumberAlt: val.phoneNumberAlt,
     })) satisfies z.ZodType<ContactInformationState>;
 
   const parsedDataResult = personalInformationSchema.safeParse({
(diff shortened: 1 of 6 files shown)


frontend/app/routes/public/apply/$id/adult-child/contact-information.tsx
```diff
@@ -177,6 +177,8 @@ export async function action({ context: { appContainer, session }, params, reque
       ...val,
       homePostalCode: val.homeCountry && val.homePostalCode ? formatPostalCode(val.homeCountry, val.homePostalCode) : val.homePostalCode,
       mailingPostalCode: val.mailingCountry && val.mailingPostalCode ? formatPostalCode(val.mailingCountry, val.mailingPostalCode) : val.mailingPostalCode,
+      phoneNumber: val.phoneNumber,
+      phoneNumberAlt: val.phoneNumberAlt,
```
sebastien-comeau, line 181: I wouldn't think this would be an issue since `val` is spreaded before. If `phoneNumber` and `phoneNumberAlt` are present then they should copied over. I will have to debug it locally.

frontend/app/routes/public/renew/$id/adult-child/confirm-phone.tsx
```diff
@@ -101,6 +101,8 @@ export async function action({ context: { appContainer, session }, params, reque
     })
     .transform((val) => ({
       isNewOrUpdatedPhoneNumber: val.isNewOrUpdatedPhoneNumber === AddOrUpdatePhoneOption.Yes,
+      phoneNumber: val.phoneNumber,
+      phoneNumberAlt: val.phoneNumberAlt,
```
sebastien-comeau, line 105: Spreading `val` should include them automatically.
```ts
.transform((val) => ({
  ...val,
  isNewOrUpdatedPhoneNumber: val.isNewOrUpdatedPhoneNumber === AddOrUpdatePhoneOption.Yes,
})
```

frontend/app/routes/public/apply/$id/adult-child/contact-information.tsx
sebastien-comeau: I've tested and it works where `val` is spreaded.

frontend/app/routes/public/renew/$id/adult-child/confirm-phone.tsx
sebastien-comeau: Spreading `val` will fix the issue, no need to explicitly add phone number fields in `.transform()`.

sebastien-comeau (approved):

PR #3075, opened 2025-01-23
AB#15698 -- Added back button link for dental insurance if applicant is reviewed

AdamP-HRSDC-RHDCC:
AB#15698 - https://dev.azure.com/ESDCCM/CDCP/_workitems/edit/15698/

### DIFF
--- a/frontend/app/routes/protected/renew/$id/member-selection.tsx
+++ b/frontend/app/routes/protected/renew/$id/member-selection.tsx
@@ -153,7 +153,7 @@ export default function ProtectedRenewMemberSelection() {
           </LoadingButton>
           <ButtonLink
             id="back-button"
-            routeId={previouslyReviewed ? "protected/renew/$id/dental-insurance" : "protected/renew/$id/tax-filing"}
+            routeId={previouslyReviewed ? 'protected/renew/$id/dental-insurance' : 'protected/renew/$id/tax-filing'}
             params={params}
             disabled={isSubmitting}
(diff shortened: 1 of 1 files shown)


frontend/app/routes/protected/renew/$id/member-selection.tsx
```diff
@@ -153,7 +153,7 @@ export default function ProtectedRenewMemberSelection() {
           </LoadingButton>
           <ButtonLink
             id="back-button"
-            routeId="protected/renew/$id/tax-filing"
+            routeId={previouslyReviewed ? 'protected/renew/$id/dental-insurance' : 'protected/renew/$id/tax-filing'}
```
nicholas-ly, line 156: We should also consider the scenario where the applicant is an ITA client and "previously reviewed". In that case the back button should take the user to `/confirm-federal-provincial-territorial-benefits`.

nicholas-ly: After reviewing the bug, I believe the redirect should only go back to `/tax-filing`. The `/member-selection` page serves as the main entry point for editing multiple members, similar to the `/children` index route (in the Adult/Child, Child public flows).

In the `/children` index route, we don’t track which child was last edited to adjust the back redirect. Instead, we navigate the user to the previous primary step.

AdamP-HRSDC-RHDCC:
Closed PR until flows are finalized

PR #3110, opened 2025-01-29
Removing underline style from non-link elements

nibivi77:
[AB#5280](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/5280)

### Additional Notes
This does not follow the Figma anymore.

All other `underline` instances in our app are for links.

An alternative could be to `hover:underline` instead of removing it completely?


sebastien-comeau:
> An alternative could be to `hover:underline` instead of removing it completely?

Based on GCWeb, `hover:underline` should work.
https://wet-boew.github.io/GCWeb/index-en.html

image

Fbasham (approved):
should be fine with this approach.

PR #3150, opened 2025-02-05
AB#5296 -- Added screenreader announcement when removing children

AdamP-HRSDC-RHDCC:
### Description
Added screen reader announcement when removing children from the children index page (all flows)

New CSS class 'visually-hidden' is a copy of MUI's visuallyHidden class. (Added in as a custom css class since we do not use MUI libraries)

[AB#5296](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/5296)

sr

### DIFF
--- a/frontend/app/tailwind.css
+++ b/frontend/app/tailwind.css
@@ -377,17 +377,3 @@
     width: auto;
   }
 }
-
-@layer utilities {
-  .visually-hidden {
-    position: absolute !important;
-    width: 1px !important;
-    height: 1px !important;
-    padding: 0 !important;
-    margin: -1px !important;
-    overflow: hidden !important;
-    clip: rect(0, 0, 0, 0) !important;
-    white-space: nowrap !important;
-    border: 0 !important;
-  }
-}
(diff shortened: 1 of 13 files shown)


frontend/app/tailwind.css
```diff
@@ -377,3 +377,17 @@
...
+    position: absolute !important;
+    width: 1px !important;
+    height: 1px !important;
+    padding: 0 !important;
+    margin: -1px !important;
+    overflow: hidden !important;
+    clip: rect(0, 0, 0, 0) !important;
+    white-space: nowrap !important;
+    border: 0 !important;
+  }
```
nibivi77, line 392: Instead of creating this css utility, we can use `className="sr-only"`

Let me know if this css utility was created for a specific reason.

nibivi77 (changes requested):

frontend/app/tailwind.css
AdamP-HRSDC-RHDCC: Switched to using the pre-existing css class 'sr-only'

PR #3176, opened 2025-02-10
Removing h2 inside <a>

nibivi77:
### Description
Removing header tag from inside anchor tags.
Small focus styling to improve visual experience.

AB#5294


Fbasham (approved):
/thumb-up

sebastien-comeau:
> it was flagged during the a11y audit.

I know but I would have check if switching `Heading` with `Link` would have been good. I don't have the whole page context. If there is already a `Heading` over the listing of the card than it should be good the way @nibivi77 changed it.

sebastien-comeau:
@Fbasham, if you want the `Heading` to remain as a heading but make the entire card clickable as a navigation link, you can use the `:after` pseudo-class on the link within the heading. This would make the pseudo-element cover the entire card. Check out the "Services and information" section on [[gov.uk](https://www.gov.uk/)](https://www.gov.uk/), where the items are "cards" with an `h3` and an `a` inside. The `a:after` is styled to cover the entire `card` container.

image

PR #3191, opened 2025-02-11
feat(frontend): add application year caching

gregory-j-baker:
### Description

Add a 24h cache for the application year lookup to reduce the number of API calls that are made.

### Checklist

### Additional nodes

Prettier formatted the code which makes it look like there are more changes than there actually are. 🤷


frontend/app/.server/domain/services/application-year.service.ts
```diff
@@ -52,9 +53,29 @@ export class DefaultApplicationYearService implements ApplicationYearService {
   }
 
   private init(): void {
+    this.log.debug('Cache TTL value: %d ms', 1000 * this.serverConfig.LOOKUP_SVC_APPLICATION_YEAR_CACHE_TTL_SECONDS);
+
+    this.findRenewalApplicationYear = moize(this.findRenewalApplicationYear, {
+      matchesKey: (cacheKey, key) => cacheKey.every((item, idx) => item.date === key[idx].date),
```
sebastien-comeau, line 59: What the following line do?

gregory-j-baker: It caches according to the current date, ignoring the passed-in user id..

I left a `TODO` suggesting we remove the user id since it isn't required, but that change would be too big for this PR.

PR #3203, opened 2025-02-12
switch InputField to InputPatternField in /stub-login

Fbasham:
### Description
This is for better quality of life/DX when using the stub-login feature.

### Checklist


nicholas-ly (commented):
I had intentionally left out the formatting of the SIN because I would sometimes use the stub login to call the Interop APIs, which accepts SINs without the spaces in their requests. We would also need to change the mocks so that the SINs are formatted too.

We could add logic in the action to remove spaces when saving on the server side if needed.

Fbasham:
> I had intentionally left out the formatting of the SIN because I would sometimes use the stub login to call the Interop APIs, which accepts SINs without the spaces in their requests. We would also need to change the mocks so that the SINs are formatted too.
> 
> We could add logic in the action to remove spaces when saving on the server side if needed.

Ahh, closing for now

PR #3218, opened 2025-02-13
fix(frontend): winston loggers

gregory-j-baker:
### Description

Hopefully fixes [AB#5330](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/5330) (ERROR: too many open files).

The core problem is that the logging transports were being re-created for every logger, instead of being created once and reused.

- [AB#5330](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/5330)


sebastien-comeau (approved):
Great!

PR #3329, opened 2025-03-11
Updated adult-child/child review adult information

AdamP-HRSDC-RHDCC:
### Description
Updated adult-child/child review adult information

[AB#5256](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/5256)
[AB#5236](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/5236)


frontend/app/routes/public/apply/$id/adult-child/review-adult-information.tsx
```diff
@@ -242,6 +244,21 @@ export default function ReviewInformation({ loaderData, params }: Route.Componen
...
+                  </>
+                ) : (
+                  <p>{t('apply-adult-child:review-adult-information.no')}</p>
+                )}
+                <div className="mt-4">
+                  <InlineLink id="change-martial-status" routeId="public/apply/$id/adult-child/applicant-information" params={params}>
+                    {t('apply-adult-child:review-adult-information.previously-enrolled-change')}
+                  </InlineLink>
+                </div>
+              </DescriptionListItem>
```
nicholas-ly, line 261: I missed this in the previous PR for the adult flow, but I think we should only display this block if `userInfo.previouslyEnrolled` is defined. The "Change past enrolment" could then redirect to the `/new-or-existing-member` route without causing any state issues.

(same comment for the `review-adult-information` in the `child` and `adult` flows)

nicholas-ly (commented):
We should also take a look at the references to dead routes. For example, there are still redirects to `/contact-information` even though this route has been removed.

frontend/app/routes/public/apply/$id/adult-child/review-adult-information.tsx
AdamP-HRSDC-RHDCC: Added conditional rendering for the block

nicholas-ly (approved):

PR #3363, opened 2025-03-18
Adult-child redirects.

nibivi77:
### Description
`verify-email` page is still missing.
All flows should work except for `verify-email`


nicholas-ly (commented):
The redirects in this PR look good and work as expected.

Something is off in the route helper on the Review page though. When I am on the `/adult-child/children` route and have a valid child added, then click "Continue with application", I get redirected to `/adult-child/parent-or-guardian`.

nicholas-ly (approved):

nicholas-ly (approved):

PR #3381, opened 2025-03-20
Following the designers intended workflow editMode in apply/adult.

nibivi77:
[AB#5426](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/5426)

### Checklist

### Test Instructions
Use a date of birth that does not take the user to the "New or existing member" screen, like Jan 1, 1970.
Proceed to the Review screen and edit the date of birth to something that will redirect the user to the "New or existing member" screen, like Jan 1, 2006 and hit "Save".
On the "New or existing member" screen and hit "Back".

It should bring you back to where you came from, while staying in editMode.

(Also works with 2008 -- living-independently)

### DIFF
--- a/frontend/app/.server/routes/helpers/apply-route-helpers.ts
+++ b/frontend/app/.server/routes/helpers/apply-route-helpers.ts
@@ -17,6 +17,13 @@ export type ApplyState = ReadonlyDeep<{
   id: string;
   editMode: boolean;
   lastUpdatedOn: string;
+  editModeApplicantInformation?: {
+    firstName: string;
+    lastName: string;
+    dateOfBirth: string;
+    socialInsuranceNumber: string;
+  };
+  editModeLivingIndependently?: boolean;
   applicantInformation?: {
     firstName: string;
     lastName: string;
--- a/frontend/app/routes/public/apply/$id/adult/applicant-information.tsx
+++ b/frontend/app/routes/public/apply/$id/adult/applicant-information.tsx
@@ -141,24 +141,44 @@ export async function action({ context: { appContainer, session }, params, reque
     return data({ errors: transformFlattenedError(parsedDataResult.error.flatten()) }, { status: 400 });
   }
 
-  saveApplyState({
-    params,
-    session,
-    state: {
-      applicantInformation: {
-        firstName: parsedDataResult.data.firstName,
-        lastName: parsedDataResult.data.lastName,
-        dateOfBirth: parsedDataResult.data.dateOfBirth,
-        socialInsuranceNumber: parsedDataResult.data.socialInsuranceNumber,
-      },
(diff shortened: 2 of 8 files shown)


nicholas-ly (commented):
FYI the scenario from the ADO still causes the date of birth to be saved even even when I cancel the update.

I wonder if we will have to introduce an extra state object to hold what the applicant initially entered - we do that for the renewal flow for addresses.

Also, if we were to be consistent with the rest of the application, when in Edit mode, I think the "Cancel" button always appears rather than "Back" (unlike on the "Living independently" and "New or existing member" screens). For renewals, when we hit "Cancel" after editing, any previous depending updates are reset back to the initial state (see address and FPT)

nibivi77:
> FYI the scenario from the ADO still causes the date of birth to be saved even even when I cancel the update.
> 
> I wonder if we will have to introduce an extra state object to hold what the applicant initially entered - we do that for the renewal flow for addresses.
> 
> Also, if we were to be consistent with the rest of the application, when in Edit mode, I think the "Cancel" button always appears rather than "Back" (unlike on the "Living independently" and "New or existing member" screens). For renewals, when we hit "Cancel" after editing, any previous depending updates are reset back to the initial state (see address and FPT)

I see what you are saying, but I don't know if we should apply what you are saying here. If the user changes their date of birth, I think they have to be redirected to "living-independently" and/or "new or existing user". In which case "Back" buttons would be appropriate in my opinion.

Same thing for the "Save" on "applicant-information". I would think that once you hit save, it saves it, even if you press "back" afterwards. Worst case you just manually change it again after they have pressed "back".

nicholas-ly:
> > FYI the scenario from the ADO still causes the date of birth to be saved even even when I cancel the update.
> > I wonder if we will have to introduce an extra state object to hold what the applicant initially entered - we do that for the renewal flow for addresses.
> > Also, if we were to be consistent with the rest of the application, when in Edit mode, I think the "Cancel" button always appears rather than "Back" (unlike on the "Living independently" and "New or existing member" screens). For renewals, when we hit "Cancel" after editing, any previous depending updates are reset back to the initial state (see address and FPT)
> 
> I see what you are saying, but I don't know if we should apply what you are saying here. If the user changes their date of birth, I think they have to be redirected to "living-independently" and/or "new or existing user". In which case "Back" buttons would be appropriate in my opinion.
> 
> Same thing for the "Save" on "applicant-information". I would think that once you hit save, it saves it, even if you press "back" afterwards. Worst case you just manually change it again after they have pressed "back".

As per designers,  we should just display Save/Cancel where "Cancel" would cancel any changes made.

nicholas-ly (approved):
Looks good to me. Let's apply this to the other flows.

PR #3416, opened 2025-03-27
AB#5429 -- Refactored confirm dental benefits save state 

AdamP-HRSDC-RHDCC:
### Description
Fixed confirm dental benefits save state

moved assignment of empty dentalBenefits object from the to the route helpers for each flow

tested on all flows and toggling the hasBenefits state page does populate/depopulate the dentalBenefits object correctly (seen in the payload at review time)

[AB#5429](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/5429)

### DIFF
--- a/frontend/app/.server/routes/helpers/apply-adult-route-helpers.ts
+++ b/frontend/app/.server/routes/helpers/apply-adult-route-helpers.ts
@@ -165,9 +165,12 @@ export function validateApplyAdultStateForReview({ params, state }: ValidateAppl
     throw redirect(getPathById('public/apply/$id/adult/dental-insurance', params));
   }
 
-  //TODO: refactor to allow dentalBenefits to be undefined if hasFederalProvincialTerritorialBenefits is false
-  if (dentalBenefits === undefined || hasFederalProvincialTerritorialBenefits === undefined) {
+  if (hasFederalProvincialTerritorialBenefits === undefined) {
     throw redirect(getPathById('public/apply/$id/adult/confirm-federal-provincial-territorial-benefits', params));
+  }
+
+  if (dentalBenefits === undefined && hasFederalProvincialTerritorialBenefits === true) {
+    throw redirect(getPathById('public/apply/$id/adult/federal-provincial-territorial-benefits', params));
   }
 
   return {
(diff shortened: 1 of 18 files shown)


frontend/app/.server/routes/helpers/apply-adult-route-helpers.ts
```diff
@@ -165,19 +165,38 @@ export function validateApplyAdultStateForReview({ params, state }: ValidateAppl
...
+    dentalBenefits:
+      hasFederalProvincialTerritorialBenefits === false
+        ? {
+            hasFederalBenefits: false,
+            federalSocialProgram: undefined,
+            hasProvincialTerritorialBenefits: false,
+            provincialTerritorialSocialProgram: undefined,
+            province: undefined,
+          }
+        : dentalBenefits,
```
nicholas-ly, line 199: We can remove the commented out.

I also feel that the route helper shouldn't be responsible for resetting the fields - the responsibility should go on the route itself, closer to where the implementation occurs.

frontend/app/routes/public/apply/$id/adult/review-information.tsx
```diff
@@ -135,7 +137,7 @@ export async function loader({ context: { appContainer, session }, params, reque
   const viewPayloadEnabled = ENABLED_FEATURES.includes('view-payload');
   const benefitApplicationDtoMapper = appContainer.get(TYPES.domain.mappers.BenefitApplicationDtoMapper);
   const benefitApplicationStateMapper = appContainer.get(TYPES.routes.mappers.BenefitApplicationStateMapper);
-  const payload = viewPayloadEnabled && benefitApplicationDtoMapper.mapBenefitApplicationDtoToBenefitApplicationRequestEntity(benefitApplicationStateMapper.mapApplyAdultStateToBenefitApplicationDto(state));
+  const payload = viewPayloadEnabled && benefitApplicationDtoMapper.mapBenefitApplicationDtoToBenefitApplicationRequestEntity(benefitApplicationStateMapper.mapApplyAdultStateToBenefitApplicationDto(state as ApplyAdultState));
```
nicholas-ly, line 140: We should remove the use of `as ApplyAdultState` here, which can mask potential typing issues. 
Instead, we can change the state mapper and make `dentalBenefits` optional (the type issue I saw when running this branch and removing `as ApplyAdultState`).

nicholas-ly (commented):
Your branch does set the payload correctly. 

However, I noticed a minor issue where the form is pre-filled when I feel it shouldn't be.
1. Fill out FPT benefits (select "Yes" and specify the FPT benefits on the next page)
2. On the Review page, edit FPT benefits and choose "No" (on the single question page)
3. On the Review page, edit FPT benefits and choose "Yes". The initial FPT benefits specified in step 1 are there. This should be cleared since we previously changed our answer to "No"

I pushed the following: https://github.com/DTS-STN/canadian-dental-care-plan/commit/e3783dd85ae3ab9d94432db31eb92c528ef3f469, which is based on what we have for renewals (for the adult flow only), which seems to address this issue as well as some of the comments I made in this PR.

Feel free to apply and use for the other flows if you see there are no issues.

frontend/app/routes/public/apply/$id/adult/review-information.tsx
AdamP-HRSDC-RHDCC: Previously I tried to use it without casting, but they way the "children: Required<ChildState>[];" is setup for the State in the Benefits Mappers removes the ability for dentalBenefis to be undefined. Removing Required<> to mitigate this causes issue across the entire benefits mappers and would require a refactoring of this mapper. I'll attempt to see if there is any other way to get this value to be undefined type acceptable. (Adult state is not an issue here, only child state objects cause this issue)

frontend/app/.server/routes/helpers/apply-adult-route-helpers.ts
AdamP-HRSDC-RHDCC: Removed commented code, and refactored resetting the fields to the example in the branch. Tested on adult/child flows and confirmed working (removes dentalBenefits on "No" answer).

frontend/app/routes/public/apply/$id/adult/review-information.tsx
AdamP-HRSDC-RHDCC: Refactored dentalBenefits setting to the confirm-fpt pages

nicholas-ly (approved):

PR #3418, opened 2025-03-28
Reworking editMode redirects in apply/adult.

nibivi77:
### Description
I think we can refactor some of this code. Its 9pm right now so I will look at this tomorrow, but it should work 'as is' for the adult flow.

[AB#5479](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/5479)


nicholas-ly (approved):
Tested and works - feel free to apply to other flows

PR #3466, opened 2025-04-03
Fixing missing i18n mocks and `DialogContent` warning in unit tests

nicholas-ly:
### Description
- `Warning: Missing `Description` or `aria-describedby={undefined}` for {DialogContent}.`
- `react-i18next:: useTranslation: You will need to pass in an i18next instance by using initReactI18next { code: 'NO_I18NEXT_INSTANCE' }`

### Checklist


gregory-j-baker (commented):
Is it worth taking a similar approach as future-sir?
<https://github.com/DTS-STN/future-sir/blob/main/frontend/__mocks__/react-i18next.tsx>

Basically, i18next-react is always mocked as

```ts
export const useTranslation = vi.fn(() => ({
  i18n: { getFixedT: () => tFunction },
  t: tFunction,
}))
```

nicholas-ly:
> Is it worth taking a similar approach as future-sir? https://github.com/DTS-STN/future-sir/blob/main/frontend/__mocks__/react-i18next.tsx
> 
> Basically, i18next-react is always mocked as
> 
> ```ts
> export const useTranslation = vi.fn(() => ({
>   i18n: { getFixedT: () => tFunction },
>   t: tFunction,
> }))
> ```

Fixed

gregory-j-baker (approved):

PR #3441, opened 2025-04-02
feat(frontend): implement application killswitch

gregory-j-baker:
### Description

Incremental PR to begin adding an application killswitch.

PR #3537, opened 2025-04-11
feat(frontend): add handling for HTTP 429 Too Many Requests across repositories

gregory-j-baker:
### Description

This is an incremental PR that cherry picks some changes from #3441 so future PRs will not be as large. This PR adds a new error code that will be thrown when an HTTP429 response is returned from an API call.

This PR doesn't alter functionality in any way, other than changing a thrown `Error` to a thrown `AppError` (for 429 responses only).


nicholas-ly (commented):
FYI `address-validation.repository` also calls Interop

PR #3543, opened 2025-04-11
Add protected child audit.

nibivi77:
### Description
As per title.

- For the files under our `child/children`, I wrote the audit with `children` instead of `child`. Should it be `child`?
- For pages where no actions are required, like `exit-application` , `cannot-apply-child`, and `parent-or-guardian`, should there be an audit in the `action`?

[AB#5580](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/5580)
[AB#5576](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/5576)


gregory-j-baker (approved):

gregory-j-baker (approved):

nicholas-ly (commented):
> For the files under our child/children, I wrote the audit with children instead of child. Should it be child?

I feel something like `view-page.apply.child.children.information` would make more sense in order to distinguish the `child` flow with `adult-child` (which would be `view-page.apply.adult-child.children.information`)

PR #3561, opened 2025-04-14
refactor(frontend) remove redundant route-helper calls in routes

Fbasham:
### Description
This PR removes redundant calls to route-helper functions in some of the routes across the project.  Generally speaking, if we don't assign a variable to a `load[flowName]...`, or we call `save[flowName]...` without any updated state, it's a redundant operation and can be removed.

### Checklist

### DIFF
--- a/frontend/app/routes/protected/apply/$id/adult-child/confirmation.tsx
+++ b/frontend/app/routes/protected/apply/$id/adult-child/confirmation.tsx
@@ -183,7 +183,6 @@ export async function action({ context: { appContainer, session }, params, reque
 
   const t = await getFixedT(request, handle.i18nNamespaces);
 
-  loadProtectedApplyAdultChildState({ params, request, session });
   clearProtectedApplyState({ params, session });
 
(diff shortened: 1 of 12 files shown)


sebcomeau (commented):
The state is loaded to check if it exists, and if it doesn't, a redirect is triggered. It also handles the 404 error, even though the return value isn't used. This functionality will be managed in the loaders and actions of all routes once the React Router middleware is available.

sebcomeau:
> @sebcomeau I see. In some cases, we call `load[flowName]...` followed by `clear[flowName]...`, but the clear function calls the load function, which is definitely redundant, no? Also the `save[flowName]` called with an argument of `{state: {}}` is redundant since the save function calls the load function, no?

I see your points. I don’t think calling `load`, even though it’s invoked in the subsequent `clear` or `save` functions, is a problem. It actually clarifies that we need to load/check if the state exists. Maybe we should update `clear` and `save` to accept the state `id` as an argument, so they can load the state and ensure the pages call `load`.

PR #3580, opened 2025-04-16
feat(frontend): Adding Express middleware to count route requests

nicholas-ly:
### Description
This PR adds a new `routeRequestCounter` Express middleware that:
- Uses `matchRoutes` to resolve the route for each request
- Caches matched route IDs by normalized path for performance
- Strips dynamic `$params` from metric/route IDs (e.g., `public.apply.$id.review` → `public.apply.review`)
- Sends a count metric to `InstrumentationService` using a route-based key like `public.apply.review.gets` to follow existing application metric conventions

### Screenshots
image

**Dynatrace:**
image

### Checklist

### Additional notes
Credit to ~~@sebastien-comeau~~ @sebcomeau for previous review and implementation

### DIFF
--- a/frontend/app/.server/express-server/instrumentation.server.ts
+++ b/frontend/app/.server/express-server/instrumentation.server.ts
@@ -1,0 +1,73 @@
+import { matchRoutes } from 'react-router';
+
+import type { RequestHandler } from 'express';
+
+import { getAppContainerProvider } from '~/.server/app.container';
+import { TYPES } from '~/.server/constants';
+import { createLogger } from '~/.server/logging';
+import routes from '~/routes';
+
+// Define a type for the cache value (can be string or null/undefined if no match/id)
+type CachedRouteId = string | null | undefined;
+
+/**
+ * Creates an Express middleware to count requests based on matched React Router routes.
+ * It caches the result of route matching for performance.
+ *
+ * @returns Express RequestHandler middleware.
+ */
+export function routeRequestCounter(): RequestHandler {
+  const log = createLogger('express.server/routeRequestCounter');
+
+  const appContainer = getAppContainerProvider();
+  const instrumentationService = appContainer.get(TYPES.observability.InstrumentationService);
+
+  // Cache to store: normalizedPath -> routeId (or null/undefined if no match)
+  // This Map persists across requests for this middleware instance.
+  const pathCache = new Map<string, CachedRouteId>();
+
+  log.info('Initializing request counter middleware with route matching cache.');
+
+  return (req, res, next) => {
+    // Hook into the 'finish' event to capture the final status code
+    res.on('finish', () => {
+      try {
+        // Use req.path to ignore query strings for routing/caching purposes
+        // Normalize path (e.g., remove trailing .data for React Router loaders/actions)
+        const normalizedPath = req.path.replace(/\.data$/, '');
+
+        let routeId: CachedRouteId = undefined;
+
+        // Check cache first
+        if (pathCache.has(normalizedPath)) {
+          routeId = pathCache.get(normalizedPath);
+          log.debug(`Cache hit for path: ${normalizedPath}, routeId: ${routeId}`);
+        } else {
+          log.debug(`Cache miss for path: ${normalizedPath}. Matching routes...`);
+
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          const matches = matchRoutes(routes as any, normalizedPath);
+
+          // Get the ID from the most specific matched route (last in the array)
+          // Ensure the route and ID exist
+          const lastMatch = matches?.at(-1); // Get the last match object
+          routeId = lastMatch?.route.id ?? null; // Use null if no ID
+
+          // Update cache with the result (even if null)
+          pathCache.set(normalizedPath, routeId);
+          log.debug(`Cached routeId '${routeId}' for path: ${normalizedPath}`);
+        }
+
+        if (routeId) {
+          // Construct metric identifier (e.g., POST '/user/$id/profile' → 'user._id.profile.posts')
+          const metricPrefix = `${routeId.replaceAll('/', '.').replace(/.\$([^.]+)/g, '_$1')}.${req.method.toLowerCase()}s`;
+          instrumentationService.countHttpStatus(metricPrefix, res.statusCode);
+        }
+      } catch (error) {
+        log.error('Error during request counting in "finish" handler:', error);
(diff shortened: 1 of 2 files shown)


frontend/app/.server/express-server/instrumentation.server.ts
```diff
@@ -0,0 +1,75 @@
...
+
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          const matches = matchRoutes(routes as any, normalizedPath);
+
+          // Get the ID from the most specific matched route (last in the array)
+          // Ensure the route and id exist and id is a string
+          const lastMatch = matches?.at(-1); // Get the last match object
+          const foundId = typeof lastMatch?.route.id === 'string' ? lastMatch.route.id : null; // Use null if no ID or not a string
+
+          routeId = foundId;
```
sebcomeau, line 56: `foundId` is unnecessary

```suggestion
          const routeId = typeof lastMatch?.route.id === 'string' ? lastMatch.route.id : null; // Use null if no ID or not a string
```

sebcomeau (changes requested):

frontend/app/.server/express-server/instrumentation.server.ts
nicholas-ly: [Fixed](https://github.com/DTS-STN/canadian-dental-care-plan/compare/d6a2cb544f04d25cb563fa3214856e4c76188a38..695f5fcb4dd566ce3f2aa4b379ff7aa6d11e7d84)

frontend/app/.server/express-server/instrumentation.server.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  log.info('Initializing request counter middleware with route matching cache.');
+
+  return (req, res, next) => {
+    // Hook into the 'finish' event to capture the final status code
+    res.on('finish', () => {
+      try {
+        // Use req.path to ignore query strings for routing/caching purposes
+        // Normalize path (e.g., remove trailing .data for React Router loaders/actions)
+        const normalizedPath = req.path.replace(/\.data$/, '');
```
gregory-j-baker, line 37: Why not leave the `.data` in the path? Might it be useful to distinguish between document vs XHR requests?

frontend/app/.server/express-server/instrumentation.server.ts
```diff
@@ -0,0 +1,73 @@
...
+        } else {
+          log.debug(`Cache miss for path: ${normalizedPath}. Matching routes...`);
+
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          const matches = matchRoutes(routes as any, normalizedPath);
+
+          // Get the ID from the most specific matched route (last in the array)
+          // Ensure the route and id exist and id is a string
+          const lastMatch = matches?.at(-1); // Get the last match object
+          routeId = typeof lastMatch?.route.id === 'string' ? lastMatch.route.id : null; // Use null if no ID or not a string
```
gregory-j-baker, line 54: Can you just do this here?

``` ts
routeId = lastMatch?.route.id ?? null;
```

frontend/app/.server/express-server/instrumentation.server.ts
```diff
@@ -0,0 +1,73 @@
...
+
+          // Update cache with the result (even if null)
+          pathCache.set(normalizedPath, routeId);
+          log.debug(`Cached routeId '${routeId}' for path: ${normalizedPath}`);
+        }
+
+        if (routeId) {
+          // Construct metric identifier (e.g., POST '/user/$id/profile' → 'user.profile.posts')
+          const metricPrefix = `${routeId.replaceAll('/', '.').replace(/\.\$[^.]+/g, '')}.${req.method.toLowerCase()}s`;
```
gregory-j-baker, line 63: Any reason the path params are removed? Why not just keep them (but normalize them to match the desired metrics key convention)?

nicholas-ly:
Marking as draft to explore Open Telemetry's auto instrumentation library.

frontend/app/.server/express-server/instrumentation.server.ts
sebcomeau: You`re right since we trust type-safety.

frontend/app/.server/express-server/instrumentation.server.ts
sebcomeau: I think @nicholas-ly want to keep current metric prefix, but I think keeping it would be better as you're suggesting.

frontend/app/.server/express-server/instrumentation.server.ts
sebcomeau: We have to remove `.path` from url pathname when calling `matchRoutes`, otherwise it cannot match URLs with `.path`.

sebcomeau: React router match routes [logic](https://github.com/remix-run/react-router/blob/ad363e148222f75d5239aae514335a1cf1f7533e/packages/react-router/lib/server-runtime/server.ts#L196) when pathname has `.data`.

nicholas-ly: The request path without `.data` is used to perform the route matching to grab the route ID, which is used as the metric ID to match existing conventions. We could use the `.data` as the metric identifier I guess..

nicholas-ly:
> Marking as draft to explore Open Telemetry's auto instrumentation library.

image
:shrug:

frontend/app/.server/express-server/instrumentation.server.ts
sebcomeau: The route ID and the pathname are distinct concepts—the route ID doesn’t need to match the pathname. We remove the `path` from the request URL's pathname to ensure `matchRoutes` can still find a match during XHR requests, since React Router appends `.data` to the pathname in those cases.

sebcomeau:
```ts
let normalizedPath = req.path;

// [url].data path are client side navigation browser requests
if (normalizedPath.endsWith('.data')) {
  normalizedPath = normalizedPath.replace(/\.data$/, '');
}
```

nicholas-ly: :thinking:

gregory-j-baker: 🤔

gregory-j-baker:
> 🤷

🤷

PR #3583, opened 2025-04-17
fix(frontend) un-default check hasSocialInsurance radios

Fbasham:
### Description
I could be mistaken and missed a key conversation when I was working on fsir, but I don't think it makes sense to have the children's `hasSocialInsuranceNumberValue` be coerced to `true` if the value is `undefined`, which makes it default checked when a user initially lands on the page.


gregory-j-baker:
I would approve but I'm not sure what the original intent was with the `?? true`.

nicholas-ly:
image
:shrug:

Fbasham:
I disagree with the designers, but ok lmao

gregory-j-baker:
Normally I would suggest never defaulting a radio.. but in this case, since there is a secondary field that (presumably) can't be left blank.. it might be okay.

PR #3591, opened 2025-04-17
Adding happy path e2e protected adult flow.

nibivi77:
### Description
Adding first simple e2e test for adult flow in protected.

This test case does not cover if the user is redirected to `living-independently` and `new-or-existing-user`.

It only covers `single` marital-status test case.

### DIFF
--- a/frontend/e2e/protected/apply/adult-category.spec.ts
+++ b/frontend/e2e/protected/apply/adult-category.spec.ts
@@ -60,14 +60,14 @@ test.describe('Adult category', () => {
 
     await test.step('Should navigate to mailing address page', async () => {
       await applyAdultPage.isLoaded('mailing-address');
-      await fillOutAddress({ address: '123 Fake Street', city: 'Ottawa', country: 'Canada', province: 'Ontario', postalCode: 'K1A 0B1', page, group: '[data-testid="mailing-address-fieldset"]' });
+      await fillOutAddress({ address: '123 Fake Street', city: 'Ottawa', country: 'Canada', province: 'Ontario', postalCode: 'K1A 0B1', page });
       await clickContinue(page);
       await page.getByRole('button', { name: 'Use selected address' }).click();
     });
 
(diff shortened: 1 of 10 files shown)


frontend/e2e/utils/helpers.ts
```diff
@@ -9,15 +9,13 @@ interface FillOutAddressArgs {
...
   province: string;
-  unit: string;
 }
 
 // Reusable function to fill out address
-export async function fillOutAddress({ address, city, country, group, page, postalCode, province, unit }: FillOutAddressArgs) {
-  const groupLocator = page.getByRole('group', { name: group });
+export async function fillOutAddress({ address, city, country, page, postalCode, province, group }: FillOutAddressArgs) {
+  const groupLocator = page.locator(group);
   await expect(groupLocator).toBeVisible();
```
nicholas-ly, line 17: Testing the visibility of the `group` (the `fieldset`) here seems unnecessary because it's always visible. This test was previously important due to the conditional rendering of the home address `fieldset` based on the user's choice to use the mailing address as the home address. 

Now that this conditional rendering is removed, we can also eliminate the need for a static `data-testid` attribute on the `fieldset` in both the `/home-address` and `/mailing-address` routes.

frontend/e2e/protected/apply/adult-category.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+      await clickContinue(page);
+    });
+
+    await test.step('Should navigate to email page', async () => {
+      await applyAdultPage.isLoaded('email');
+      await page.getByRole('textbox', { name: 'Email address' }).fill('test@example.com');
+      await clickContinue(page);
+    });
+
+    await test.step('Should navigate to email page', async () => {
```
nicholas-ly, line 101:
```suggestion
    await test.step('Should navigate to verify email page', async () => {
```

nicholas-ly (approved):

PR #3595, opened 2025-04-22
fix(frontend) fix a11y element cant be a child of

Fbasham:
### Description
According to [HTML validator](https://validator.w3.org/), some of our markup is invalid.  This was reported by the ITAO.

[AB#5647](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/5647)

### Checklist

### DIFF
--- a/frontend/app/components/collapsible.tsx
+++ b/frontend/app/components/collapsible.tsx
@@ -8,7 +8,7 @@ export type CollapsibleSummaryProps = ComponentProps<'summary'>;
 export function CollapsibleSummary({ children, className, ...props }: CollapsibleSummaryProps) {
   return (
     <summary className={cn('cursor-pointer marker:text-blue-900', className)} {...props}>
-      <div className="ml-4 inline-block text-blue-900 hover:underline">{children}</div>
+      <span className="ml-4 inline-block text-blue-900 hover:underline">{children}</span>
     </summary>
   );
 }
--- a/frontend/app/components/input-help.tsx
+++ b/frontend/app/components/input-help.tsx
@@ -2,7 +2,7 @@ import type { ComponentProps, ReactNode } from 'react';
 
 import { cn } from '~/utils/tw-utils';
 
-export interface InputHelpProps extends ComponentProps<'span'> {
+export interface InputHelpProps extends ComponentProps<'div'> {
   children: ReactNode;
   id: string;
 }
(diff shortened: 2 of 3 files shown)


sebcomeau:
Can you tell me in which page it causes an issue?

Fbasham:
> Can you tell me in which page it causes an issue?

anywhere we use `<Collapsible />`, `<div>`s can't be children of `<span>` or `<summary>`.  Why this wasn't caught in previous rounds of assessments?  I'm guessing because every auditor does things differently and they don't all run the HMTL validator.

frontend/app/components/collapsible.tsx
```diff
@@ -8,7 +8,7 @@ export type CollapsibleSummaryProps = ComponentProps<'summary'>;
 export function CollapsibleSummary({ children, className, ...props }: CollapsibleSummaryProps) {
   return (
     <summary className={cn('cursor-pointer marker:text-blue-900', className)} {...props}>
-      <div className="ml-4 inline-block text-blue-900 hover:underline">{children}</div>
+      <span className="ml-4 inline-block text-blue-900 hover:underline">{children}</span>
```
sebcomeau, line 11: This will cause issue since children can be anything and great chance it will have `DIV`s  as children. Keep it as `DIV`.

frontend/app/components/input-help.tsx
```diff
@@ -2,16 +2,16 @@ import type { ComponentProps, ReactNode } from 'react';
...
+export interface InputHelpProps extends ComponentProps<'div'> {
   children: ReactNode;
   id: string;
 }
 
 export function InputHelp(props: InputHelpProps) {
   const { children, className, ...restProps } = props;
   return (
-    <span className={cn('block max-w-prose text-gray-500', className)} data-testid="input-help" {...restProps}>
+    <div className={cn('block max-w-prose text-gray-500', className)} data-testid="input-help" {...restProps}>
```
sebcomeau, line 13: `block` class is not required since `DIV` is a block element by default.
```suggestion
    <div className={cn('max-w-prose text-gray-500', className)} data-testid="input-help" {...restProps}>
```

frontend/app/components/collapsible.tsx
sebcomeau: You're right. I've tested and `div` is not allowed with `summary`.

image

sebcomeau (approved):

PR #3644, opened 2025-04-28
fix(frontend) make address-validation-dialog outAriadescribedById unique

Fbasham:
### Description
per comment in slack


frontend/app/components/address-validation-dialog.tsx
sebcomeau, line 1: `outerAriaDescribedById={`verify-description-${useId()}`}` won't work since the generated ID doesn't match `DialogDescription` id attribute `"verify-description"`.

Here's my suggestion:

```tsx
const dialogDescriptionId = useId()
...

<DialogDescription id={dialogDescriptionId}>
...

outerAriaDescribedById={dialogDescriptionId}
```

sebcomeau (approved):

PR #3691, opened 2025-05-05
AB#5890 -- Fixed edit mode verify email bug

AdamP-HRSDC-RHDCC:
### Description
Fixed public/protected edit mode verify email bug where email verification in the payload would be set to false on cancel.

[AB#5890](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/5890)


nicholas-ly (commented):
This change covers most scenarios. 

However, if both communication preferences are initially set to Mail, and I then edit them, changing one or both to Email, then proceed to `/verify-email` and click Cancel, the application now sends `"ApplicantEmailVerifiedIndicator": true`, even though both communication preferences are still set to Mail.

PR #3750, opened 2025-05-14
fix(frontend) add email verification to public renew ITA app

Fbasham:
### Description
adds email verification to public renew ITA app.

NOTES: 
`ApplicantEmailVerifiedIndicator ` in the payload still needs to be fixed.  As far as I'm aware, this is still the case for the other public renew apps?  Correct me if I'm wrong.

AB#5951


nicholas-ly:
> ApplicantEmailVerifiedIndicator  in the payload still needs to be fixed. As far as I'm aware, this is still the case for the other public renew apps? Correct me if I'm wrong.

You would be correct - will add a task in existing ADO so it's not forgotten.

nicholas-ly (approved):

PR #3809, opened 2025-05-29
refactor(frontend) update letter-type mock implementation

Fbasham:
### Description
When calling interop to fetch letter-types, it was discovered that the shape of the data differs from the current data we use.  This is the first PR in a series to refactor the letter-type service.  This PR addresses refactoring the mock implementation.  Subsequent PRs will address the default implementation.  The invitation letter type ID has been made configurable so as to remove it at the service layer since this is apparently a business requirement to not display it to users.

AB#6170

### Checklist

### DIFF
--- a/frontend/.env.example
+++ b/frontend/.env.example
@@ -356,6 +356,10 @@ CLIENT_STATUS_SUCCESS_ID=51af5170-614e-ee11-be6f-000d3a09d640
 # Invalid statusId that can be returned when using status checker; must be treated like a null statusId
 INVALID_CLIENT_FRIENDLY_STATUS=504fba6e-604e-ee11-be6f-000d3a09d640
 
+# Invalid letter type IDs to filter at the service layer - comma separated string of IDs
+# (optional; default: 775170000)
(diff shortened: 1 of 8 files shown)


frontend/.env.example
```diff
@@ -356,6 +356,10 @@ CLIENT_STATUS_SUCCESS_ID=51af5170-614e-ee11-be6f-000d3a09d640
 # Invalid statusId that can be returned when using status checker; must be treated like a null statusId
 INVALID_CLIENT_FRIENDLY_STATUS=504fba6e-604e-ee11-be6f-000d3a09d640
 
+# Invitation letter type ID to filter at the service layer
+# (optional; default: 775170000)
+INVITATION_LETTER_TYPE_ID=775170000
```
nicholas-ly, line 361: It might be a good idea to define this configuration as an array containing all the letter types we want to filter - currently just `775170000`. 

This approach would give us flexibility in case additional letter types need to be filtered in the future (since we don't have control over the API), or if we discover a missing type in production. We could then update the configuration without requiring a new release.

PR #3824, opened 2025-06-03
fix(frontend) force email verification modal open when email request sent

Fbasham:
### Description
The modal was only showing in `editMode` because the status of `verification-code-sent` was only inside of that block.  This PR moves the return out one level when the request link is clicked.

AB#6212

### DIFF
--- a/frontend/app/routes/protected/renew/$id/verify-email.tsx
+++ b/frontend/app/routes/protected/renew/$id/verify-email.tsx
@@ -105,9 +105,16 @@ export async function action({ context: { appContainer, session }, params, reque
 
     if (state.editMode) {
       invariant(state.editModeEmail, 'Expected editModeEmail to be defined');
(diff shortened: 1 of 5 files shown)


frontend/app/routes/protected/renew/$id/verify-email.tsx
```diff
@@ -112,8 +112,8 @@ export async function action({ context: { appContainer, session }, params, reque
         preferredLanguage: state.clientApplication.communicationPreferences.preferredLanguage === ENGLISH_LANGUAGE_CODE.toString() ? 'en' : 'fr',
         userId: idToken.sub,
       });
-      return { status: 'verification-code-sent' } as const;
     }
+    return { status: 'verification-code-sent' } as const;
```
nicholas-ly, line 116: Although moving this line would display the dialog, the verification code won't actually be emailed unless the user is in `editMode`.

nicholas-ly (approved):

PR #3844, opened 2025-06-09
refactor(frontend) preferred communication code tables

Fbasham:
### Description
refactors our code table to use an internal id and then maps to the PP specific identifiers upon the final request.

AB#6319

### Checklist


nicholas-ly (approved):
Approving since the changes work and internal ids are no longer exposed to the client.

However, we can refactor this further. The repository's role is to interact with downstream APIs or data stores, which won't be the case for "preferred communication methods". We should consider removing the repository and service layers entirely and instead, define the values exposed to the client directly in the route.

This would be similar to how we avoid using a service/repository for the static yes/no values here: https://github.com/DTS-STN/canadian-dental-care-plan/blob/fac888f0048bab8b4bb77aa317be3dfdbce7ee06/frontend/app/routes/public/apply/%24id/adult/federal-provincial-territorial-benefits.tsx#L42

PR #3853, opened 2025-06-11
AB#6319 -- Updated method of communication code table to use string internal id

AdamP-HRSDC-RHDCC:
### Description
Updated method of communication code table to use string internal id

AB#6319


AdamP-HRSDC-RHDCC:
> I was going to make a similar comment to [#3844 (review)](https://github.com/DTS-STN/canadian-dental-care-plan/pull/3844#pullrequestreview-2910992627) but just noticed that the preferred communication method service isn't being called anywhere. It looks like we aren't exposing internal ids for any of the communication method questions.
> 
> If you can confirm that's the case, we should be able to remove the entire preferred communication method services, repositories, mappers, etc., similar to #3849.

Can confirm it is only used in its own test, I will remove it and associated components.

nicholas-ly (approved):

PR #3917, opened 2025-07-03
chore(gitops) adjust letter_type_ttl_cache to 300 seconds for uat1

Fbasham:
per title


gitops/overlays/uat1/configs/frontend/config.conf
```diff
@@ -28,3 +28,5 @@ ADOBE_ANALYTICS_SRC=https://assets.adobedtm.com/be5dfd287373/9b9cb7867b5b/launch
 
 # temporarily set killswitch TTL to two minute for easier testing
 APPLICATION_KILLSWITCH_TTL_SECONDS=120
+
+LOOKUP_SVC_LETTER_TYPE_CACHE_TTL_SECONDS=300
```
nicholas-ly, line 32: I think we may need to configure `LOOKUP_SVC_ALL_LETTER_TYPES_CACHE_TTL_SECONDS` as well.

nicholas-ly (approved):

PR #4012, opened 2025-08-19
fix(frontend): Express session issue

sebcomeau:
### Description

Fix Express session issue where calling `save` was preventing new sessions from being persisted correctly. When users accessed `/en/apply` directly, they were redirected to the T\&C page with a new apply flow ID, but the state was not stored in Redis. This problem only occurred with Redis Sentinel. The fix removes explicit `save` calls on session `set` and `unset`, since the session is already saved before the response is sent. Verified locally with Redis Sentinel.

[Session.save(callback)](https://github.com/expressjs/session?tab=readme-ov-file#sessionsavecallback)

> This method is automatically called at the end of the HTTP response if the session data has been altered (though this behavior can be altered with various options in the middleware constructor). Because of this, typically this method does not need to be called.

#### How to reproduce the issue

Open new Incognito window and open this URL: https://cdcp-int1.dev-dp-internal.dts-stn.com/en/apply . You should be redirected to Canada CDCP Website. The fix can be tested after this PR is merged and deployed in INT1.

### Checklist

### DIFF
--- a/frontend/__tests__/.server/web/session.test.ts
+++ b/frontend/__tests__/.server/web/session.test.ts
@@ -101,14 +101,12 @@ describe('Session', () => {
         const session = new ExpressSession(mockRequestSession);
         session.set('key' as SessionKey, 'value');
         expect(mockRequestSession.key).toBe('value');
-        expect(mockRequestSession.save).toHaveBeenCalledOnce();
       });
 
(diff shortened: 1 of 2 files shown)


gregory-j-baker (approved):
lgtm!

PR #5340, opened 2026-08-25
fix(a11y): prevent screen readers from announcing HTML markup in name field descriptions

shawn320:
### Description
The `nameInstructions` string contains `<strong>both</strong>` markup, which
renders correctly in the visual Collapsible (via `<Trans>`), but is announced
literally by screen readers when passed directly to `aria-description`.

- Added a `stripHtml` utility in `app/.server/utils/string-utils.ts`
  that uses `jsdom` to remove HTML tags from a string.
- Both `personal-information.tsx` loaders now derive the plain-text description
  once via `stripHtml(t(($) => $.personalInformation.nameInstructions))` and
  pass it down through `loaderData`; the component reads it for
  `aria-description`.

[AB#8728](https://dev.azure.com/DTS-STN/1fc40a8f-28cf-47bc-b6e4-1c234bd06177/_workitems/edit/8728)

### Checklist


sebcomeau (dismissed):
Add a string utility function to strip HTML markups from any string instead. This will prevent content duplication. I'm using `jsdom` here because it's something that we might want to use on NodeJs server in the future.

```ts
import { JSDOM } from "jsdom";

/**
 * Removes HTML tags from a string.
 *
 * @param content - String that may contain HTML.
 * @returns The string without HTML tags.
 *
 * @example
 * stripHtml("Hello <strong>world</strong>.");
 * // "Hello world."
 */
export function stripHtml(content: string): string {
  const dom = new JSDOM("");
  const element = dom.window.document.createElement("div");
  element.innerHTML = content;
  return element.textContent ?? "";
}
```

frontend/app/routes/protected/application/spokes/personal-information.tsx
```diff
@@ -236,7 +238,7 @@ export default function ApplicationPersonalInformation({ loaderData, params }: R
                   label={t(($) => $.personalInformation.firstName)}
                   className="w-full"
                   maxLength={100}
-                  aria-description={t(($) => $.personalInformation.nameInstructions)}
+                  aria-description={nameInstructionsAriaDescription}
```
sebcomeau: Could we make `stripHtml` universal and call it directly while rendering the component? It would then work during both server-side rendering and client-side rendering. The implementation may need to account for the runtime, though: `JSDOM` is useful on the server but shouldn’t be bundled for the client, where native DOM APIs are already available.

```suggestion
                  aria-description={stripHtml(t(($) => $.personalInformation.nameInstructions))}
```

sebcomeau: Well, `stripHtml` is not needed. I will add another comment that will simplify this a11y fix.

sebcomeau (changes requested):
We should use `aria-describedby` to reference the instructions already rendered by the `Collapsible`, rather than parsing the translation into a separate string?

`Collapsible` assigns `name-instructions-content` to its content container, so both fields could reference it directly:

```suggestion
  aria-describedby="name-instructions-content"
```

This would keep a single source of content and allow the accessibility tree to derive plain text from the rendered markup. The `stripHtml` utility and its `JSDOM` dependency would no longer be needed, avoiding additional server-side processing and server/client rendering differences.

The same change would apply to the other name fields.

sebcomeau (approved):

sebcomeau (approved):

sebcomeau (commented):
All good!

PR #5362, opened 2026-09-02
fix(a11y): focus and announce newly added child on children's application (part 3)

shawn320:
### Description

Adding a child now updates the session and returns
structured action data instead of redirecting, letting the fetcher revalidate
the loader in place. After the list re-renders, focus moves to the new child's
heading and a live-region message announces "Child N was added successfully."

The operation is derived from the action's returned data — never inferred by
comparing child counts before and after.

- Add `@react-aria/live-announcer` for live-region announcements.
- Add a shared `useFetcherActionComplete` hook that runs a callback once, after
  a fetcher submission and its revalidation complete.
- `add` action now saves the child and returns `{ operation, childId, childNumber }`
  instead of redirecting to the same route.
- Give each child heading an `id` and `tabIndex={-1}` so it can receive focus.
- Add `childAddedAnnouncement` (EN/FR) translations.

### Checklist

- [ ] I have checked that my code follows the project's coding style by running `pnpm run format:check`
- [ ] I have checked that my code contains no linting errors by running `pnpm run lint`
- [ ] I have checked that my code contains no type errors by running `pnpm run typecheck`
- [ ] I have checked that all unit tests pass by running `pnpm run test:unit -- run`
- [ ] I have checked that all e2e tests pass by running `pnpm run test:e2e`

### DIFF
--- a/frontend/app/.server/locales/en/application-full-child.ts
+++ b/frontend/app/.server/locales/en/application-full-child.ts
@@ -185,6 +185,7 @@ const ns = {
     removeChild: "Remove child",
     removeChildAccessibleName: "Remove child {{childNumber}}",
     removeChildAccessibleNameWithName: "Remove child {{childNumber}}, {{childName}}",
+    childAddedAnnouncement: "Child {{childNumber}} was added successfully.",
   },
 } as const;
 
--- a/frontend/app/.server/locales/en/application-full-family.ts
+++ b/frontend/app/.server/locales/en/application-full-family.ts
@@ -205,6 +205,7 @@ const ns = {
     removeChild: "Remove child",
     removeChildAccessibleName: "Remove child {{childNumber}}",
     removeChildAccessibleNameWithName: "Remove child {{childNumber}}, {{childName}}",
+    childAddedAnnouncement: "Child {{childNumber}} was added successfully.",
   },
   exitApplication: {
     pageTitle: "Exiting the application",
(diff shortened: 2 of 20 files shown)


frontend/app/hooks/use-fetcher-action-complete.ts
```diff
@@ -0,0 +1,38 @@
...
+ * already revalidated and the updated UI has been committed. This makes the hook
+ * well suited for post-submission side effects—such as moving focus to newly
+ * rendered content or announcing a live-region message—that must not rely on
+ * inferring the operation from before/after data comparisons. The action itself
+ * returns the structured data describing what happened.
+ *
+ * @param fetcher A fetcher-like object exposing `state` and `data`.
+ * @param onComplete Callback invoked with the fetcher's returned action data.
+ */
+export function useFetcherActionComplete<TData>(fetcher: FetcherLike<TData>, onComplete: (data: NonNullable<TData>) => void): void {
```
sebcomeau, line 27: Could we store the latest `onComplete` callback in a ref and remove it from the dependency array? This would keep the completion effect tied to changes in the fetcher’s state and data rather than changes to the callback’s identity.

Here’s an example of this pattern: https://github.com/STN-DTS/SIRR-RRAS/blob/main/apps/node/agent-portal/app/hooks/use-countdown.ts#L50-L56

**Why this matters**

If callers pass a callback that isn’t memoized, its identity changes on every render and causes this effect to run unnecessarily. A ref provides access to the latest callback without making callback identity part of the effect’s lifecycle.

sebcomeau (approved):

